Issue #5, opened 2024-08-25
Document process for binding additional custom Qt libraries


mappu:
e.g. QScintilla, QZXing

PR #44, opened 2024-10-16
QScintilla (GPL) and ScintillaEdit (MIT) support

mappu:
Updates: #5 

Updates: #45 

Requires: #47

Adds bindings for QScintilla. This library is GPL/Commercial only :warning: and is put in a new `restricted-extras` section.

Adds bindings for ScintillaEdit (MIT).

**Screenshots:**


jamalsa:
I'm not an expert on Qt or C++, but is there any chance for the binding to use Scintilla component (https://www.scintilla.org) which has permissive license directly? I noticed in their source there are Qt components (ScintillaEditBase and ScintillaEdit).

mappu:
@jamalsa That's really interesting, thank you for pointing it out. ScintillaEdit builds cleanly for me, it seems like a less featureful binding than QScintilla (no signals at all) but the license is definitely preferable.

Let's add both,

PR #66, opened 2024-11-10
Add trivialwizard6

rcalixte:
This uses Qt6 QWizardPage instances to create a sample introductory wizard for an application.

This is ported from:
https://github.com/jimmykuu/therecipe-qt-examples/blob/master/widgets/dialogs/trivialwizard/trivialwizard.go

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ examples/helloworld6/helloworld6
 examples/mdoutliner/mdoutliner
 examples/windowsmanifest/windowsmanifest
 examples/uidesigner/uidesigner
+examples/trivialwizard6/trivialwizard6
 examples/libraries/extras-scintillaedit/extras-scintillaedit
 examples/libraries/qt-multimedia/qt-multimedia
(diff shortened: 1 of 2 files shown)


rcalixte:
Would it makes sense to move the examples to a dedicated repository and link to them from the README or as a git submodule? I want to flesh out the Qt6 examples so I expect to contribute more of them but it might also help other developers onboard with this toolkit.

mappu:
Nice, the example looks great.

> Would it makes sense to move the examples to a dedicated repository and link to them from the README or as a git submodule?

I'm happy either way.

- One reason to keep some examples in this repo is so that CI can compile them, to avoid breakage. This is especially important as MIQT is still not yet v1.0.0.
- One reason to split to another repo, is the ability to port examples with a wider variety of licenses. It's great to port from jimmykuu/therecipe-qt-examples, but because code in therecipe/qt is LGPL, we could only port some small trivial things to avoid copyright infringement. But in a new repo with a different license, that would definitely not be a problem.

PR #77, opened 2024-11-13
genbindings: Lint cleanup

rcalixte:
Some quick things I found while looking through for other changes. 😄


mappu:
Nice, thank you!

Issue #69, opened 2024-11-10
Connecting to Events (KeyPressEvent, WheelEvent, Move, Resize etc)


ilius:
Hi

I'm using `qt6` package.

There are several widget signals I could not figure out how to connect to (register a func to run)

- [keyPressEvent](https://doc.qt.io/qt-6/qwidget.html#keyPressEvent)
- [moveEvent](https://doc.qt.io/qt-6/qwidget.html#moveEvent)
- [resizeEvent](https://doc.qt.io/qt-6/qwidget.html#resizeEvent)
- [mouseReleaseEvent](https://doc.qt.io/qt-6/qwidget.html#mouseReleaseEvent)
- [wheelEvent](https://doc.qt.io/qt-6/qwidget.html#wheelEvent)
- [contextMenuEvent](https://doc.qt.io/qt-6/qwidget.html#contextMenuEvent)

I also could not figure out how to use `QObject.Connect2` manually for these.

mappu:
MIQT supports Qt signals well, but in C++ those examples are not signals, but rather virtual override class methods. Qt uses this technique for methods that might be called with a high frequency. 

Duplicate of #42 . Work is in progress to support this very soon.

Issue #42, opened 2024-10-13
Support subclassing


mappu:
Discussion from https://news.ycombinator.com/item?id=41785568 :

> > > Is it possible to extend/subclass an existing QWidget and customise it?
> > >
> > For e.g. PaintEvent() from pure Go code? Not yet, but I have some idea on how to make it possible.
> >
> > Until then, a nice feature of CGO is you can write one custom Qt subclass in a .cpp file in the same folder as your .go files. Your C++ will be compiled with `go build`, and you can use your custom class directly from Go/MIQT. So even if some advanced Qt feature is not in the binding, anything is still possible at least. I'll make an example about this.
>
> I contribute to NodeGui, Qt on Nodejs. It doesn't try to emulate the C++ API by allowing subclassing. Instead it exposes most methods on QWidget classes/subclasses which process QEvent objects (and their subclasses) in the form of events which you can subscribe to as an application. You can listen to signals and also QEvent related traffic in much the same way. So, if you want to subclass a QWidget to do your own custom rendering/painting, you would create an instance of QWidget and then listen to/hook into, the incoming requests to repaint. Qt sees a normal QWidget instance, but you can customise it do act like the subclass you desire. This avoids any attempt to do a real C++ subclass at runtime. Internally, NodeGui subclasses almost every QWidget (and more) to hook into them to allow this. This works far better than you might expect and saves a ton of complexity.

mappu:
[QObject::InstallEventFilter](https://doc.qt.io/qt-6/qobject.html#installEventFilter) may be an alternative to subclassing.

jamalsa:
I think we can follow therecipe/qt to expose most of qt methods. As an example to override `data` for QAbstractItemModel it has `ConnectData` which accept a go function which will be called from C++ side whenever `data` method is called. The downside is we need to generate C++ subclass for each Qt classes to handle this which will make the binding bigger and longer to compile.

jamalsa:
Thanks @mappu, I do some test on Windows with Qt 6.5.3 building `examples/helloworld6` on my  i7-8665U CPU (5 years old CPU) and 16 GB RAM 

|                         | master | miqt-subclass | change |
| ---------------- | ------------- | ------------- | ------------- |
| first build time | 00:20:28  | 00:23:47  | 16.2%  |
| binary size  | 6.470 KB  | 12.444 KB  | 92.3%  |

So the binary size almost doubled, but I think it is acceptable trade off if we can subclass Qt class.

[...]

PR #78, opened 2024-11-15
Add subclassing support

mappu:
Fixes: #42 

Fixes: #69 

Fixes: #80 

Subclass all Qt classes containing virtual methods, and add On() wrapper functions to implement the virtual override. This works for all virtual methods including paintEvent and overloaded operators.

C++ virtual subclasses have a different pointer value. Change CABI constructors to return every virtual subclass pointer so that Go types can be accurate.

Add support for calling super() in overloaded callbacks.

---

Remaining work:

- [x] Fix remaining compile issues
    - [ ] Fix defer free happening too early in `miqt_exec_callback_QAccessibleWidget_Relations`
- [X] Add super() helper to Go callback
- [X] Check if compile time is reasonable or if other mitigation is required (e.g. move every class into an optional Go subpackage)
    - Seems about 20% slower
- [x] Add example
- [x] Usage guide in the README


rcalixte:
> Sorry for the stupid question, but how should I proceed to subclass something like https://doc.qt.io/qt-6/qabstracttablemodel.html?

Not a dumb question! If you look at the library implementation, you'll see that it subclasses many of the class implementations already. So each instance you get back via Go, is already a subclass if supported. You can then use the overrides to adjust things as needed. Do you have a specific example in mind? #246 is still an issue but other than that, things should work reasonably well.

fiatjaf:
Well, basically I'm having a hard time getting a ComboBox to have a Table view and I thought that was because I had to implement a subclass of TableAbstractView (I forgot the names) and I wasn't sure how to proceed about it.

I guess the answer must be that I have to do something else but I can't figure out what.

rcalixte:
> Well, basically I'm having a hard time getting a ComboBox to have a Table view and I thought that was because I had to implement a subclass of TableAbstractView (I forgot the names) and I wasn't sure how to proceed about it.
> 
> I guess the answer must be that I have to do something else but I can't figure out what.

Do you want to open a [discussion](https://github.com/mappu/miqt/discussions) about this? This is a closed pull request. 😅

PR #82, opened 2024-11-17
Modify qt6/cbor for proper builds

rcalixte:
Tested with Debian 12 and Qt 6.4


rcalixte:
Should I include the diffs here? I can do whatever you'd like but I imagine you would prefer that they match your expectations.

mappu:
Thanks - really nice find :facepalm: 

Yes, I think generally the Qt diffs should be included in the PR (maybe as a second commit). The CI pipeline has a step that re-runs genbindings, to make sure that the bindings can always be generated cleanly with no diff.

I'll push the diffs to your branch now,

Issue #72, opened 2024-11-11
Call runtime.LockOSThread automatically in NewQApplication


mappu:
> Wow. Indeed, calling runtime.LockOSThread() before NewQApplication seems to make it always work.

_Originally posted by @ilius in https://github.com/mappu/miqt/discussions/68#discussioncomment-11202894_

ilius:
Although if `NewQApplication` itself is called in another function which is called in `main()`, this may not fix the problem (I had that issue in FreeBSD).
I think `runtime.LockOSThread()` should be called directly inside `main()`.

mappu:
@ilius Are you sure?

The documentation for LockOSThread says it should affect the entire goroutine indefinitely. It doesn't stop taking effect when returning from the function, it only stops taking effect when explicitly unlocking with UnlockOSThread.

But, real experience beats documentation.

Sorry to bother you but since I can't currently reproduce this, would you be able to remove your own calls to LockOSThread and try the branch from #83 ?

ilius:
I got this error with #84
https://gist.github.com/ilius/48a96df3d4acd5cb06757f77dcbb7ceb

Then I ran `sudo rm -rf $GOPATH/pkg/mod/github.com/mappu/miqt\@v0.*` and `go build` again, but it finished quickly.
So I ran `rm -rf ~/.cache/go-build/` and `go build` again, and I got the same error as above.

ilius:
I get this error on `master` as well.
I might have updated my Qt since last build.
How to fix it?

mappu:
> Cannot mix incompatible Qt library (6.7.3) with this library (6.7.2)

I think this is caused by a mismatched version of your qt6 and qt6-multimedia packages. Really they should have the same version. Can you check in your package manager if there are more updates?

ilius:
They are both 6.7.3, the latest version.

ilius:
FreeBSD's pkg is not as clever as Linux's apt or dnf I guess!

ilius:
Moving `runtime.LockOSThread` out of `main()` func to before `qt.NewQMainWindow` works now.
Maybe I was doing something wrong before!

Also #83 without `runtime.LockOSThread` works.

Next I removed `runtime.LockOSThread` and reverted back miqt to master, compiled and ran, this time with no error! But the app window does not show up! Did they change something in Qt 4.7.3 that caused this (it sounds worse!).

Anyway, #83 fixes it.

mappu:
[...]

PR #83, opened 2024-11-18
Call runtime.LockOSThread() automatically for the Qt main thread

mappu:
Fixes: #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ Qt class inherited types are projected as a Go embedded struct. For example, to 
 
 - When a Qt subclass adds a method overload (e.g. `QMenu::addAction(QString)` vs `QWidget::addAction(QAction*)`), the base class version is shadowed and can only be called via `myQMenu.QWidget.AddAction(QAction*)`.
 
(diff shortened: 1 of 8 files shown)


jamalsa:
Shouldn't we call `runtime.LockOSThread()` in `init` function on `qt`/`qt6` package like shown here: https://go.dev/wiki/LockOSThread? This way we don't need to have a separate call in three places.

mappu:
The most important thing is that the goroutine that calls LockOSThread is the same one that calls NewQApplication:

- For most MIQT apps, I expect NewQApplication would be called early in main, and so putting LockOSThread in init as per the wiki would have the same effect.
- But, in some cases it might be possible that NewQApplication is called a long time after init. For example, if some app is mostly a non-Qt app (e.g. a daemon) and only initializes a Qt GUI in response to some later event. In that case, LockOSThread would not match correctly.

So I think it's safer to do it this way. The code in three places is not a maintenance problem, since it's autogenerated, and it exists only in one place in the generator. Although for binary size reasons, maybe it would be nice to have only one.

PR #114, opened 2024-12-29
Add QVariant__Color support

st0pli:
Add QVariant__Color support


rcalixte:
There's a few issues here:

1. You are modifying the bindings directly rather than the codegen. That means these changes will not persist, let alone pass a CI build.
2. This functionality is [not even supported upstream](https://doc.qt.io/qt-6/qvariant.html#a-note-on-gui-types). See if you can implement what is documented there with the Go bindings and maybe open a detailed issue if you encounter any problems.

st0pli:
> There's a few issues here:
> 
> 1. You are modifying the bindings directly rather than the codegen. That means these changes will not persist, let alone pass a CI build.
> 2. This functionality is [not even supported upstream](https://doc.qt.io/qt-6/qvariant.html#a-note-on-gui-types). See if you can implement what is documented there with the Go bindings and maybe open a detailed issue if you encounter any problems.

I don't know much about QT, but I tried to implement the following functions and failed, so I modified them
Can you give me an example of how to do it
demo

```
package main
import (
	"fmt"
	"os"
	qt "github.com/mappu/miqt/qt6"
)
func main() {
	qt.NewQApplication(os.Args)
	model := qt.NewQAbstractListModel()
		return 1000
	})
	model.OnData(func(idx *qt.QModelIndex, role int) *qt.QVariant {
		if !idx.IsValid() {
			return qt.NewQVariant()
		}
		switch qt.ItemDataRole(role) {
		case qt.ForegroundRole:
			if idx.Row() % 2 == 0 {
				return qt.NewQVariant4099(qt.NewQColor3(0, 0, 0))
			} else {
				return qt.NewQVariant4099(qt.NewQColor3(255, 0, 0))
			}
		case qt.BackgroundRole:
			if idx.Row() % 2 == 0 {
				return qt.NewQVariant4099(qt.NewQColor3(255, 255, 255))
			} else {
				return qt.NewQVariant4099(qt.NewQColor3(80, 80, 80))
			}
		case qt.DisplayRole:
			return qt.NewQVariant14(fmt.Sprintf("this is row %d", idx.Row()))
		default:
			return qt.NewQVariant()
		}
	})
	v := qt.NewQListView2()
	v.SetModel(model.QAbstractItemModel)
	v.Show()
	qt.QApplication_Exec()
}
```

mappu:
In C++, this works because `QColor` has a `operator QVariant()` defined: https://doc.qt.io/qt-6/qcolor.html#operator-QVariant . Then, doing `QVariant(QColor(...))` first calls `QColor.operator QVariant()`, and then secondly calls the `QVariant` copy-constructor.

I think Miqt's bindings should have already generated some `QColor.OperatorQVariant()` function, but for some reason it didn't appear. To check further, the log output of `genbindings` should be inspected.

rcalixte:
> I think Miqt's bindings should have already generated some `QColor.OperatorQVariant()` function, but for some reason it didn't appear.

This is because it's an overloaded operator and those are skipped right?

mappu:
Yes. From checking the `cachedir/__usr__include__x86_64-linux-gnu__qt5__QtGui__qcolor.h.json` it looks like `CXXConversionDecl`. This use case seems important, so it needs to be implemented in `clang2il.go`. It should be somewhat easy.

st0pli:
I looked at the gen_qvariant.go gen_qvariant.h gen_qvariant.cpp files, and there are some that don't support it either

mappu:
Hi @st0pli ,

I can't merge this PR because the change has to be made inside the codegenerator. So, I made the change in #120. I added your example.

It will work for Font, Brush, Color, and also many more types.

st0pli:
@mappu It's working fine, it's cool

Issue #3, opened 2024-08-25
32-bit OS support


mappu:
~~Probably this works already, but it should be tested and documented.~~

mappu:
Tested 2024-12-30 with v0.7.1+

```
/src/qt/gen_qaccessible.go:192:56: cannot use 4294967295 (untyped int constant) as QAccessible__RelationFlag value in constant declaration (overflows)
/src/qt/gen_qgraphicsitem.go:107:58: cannot use 2147483648 (untyped int constant) as QGraphicsItem__Extension value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:51:42: cannot use 4261412864 (untyped int constant) as KeyboardModifier value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:61:27: cannot use 4261412864 (untyped int constant) as Modifier value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:104:32: cannot use 4294967295 (untyped int constant) as MouseButton value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:258:51: cannot use 2147483648 (untyped int constant) as WindowType value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:1308:46: cannot use 2147483648 (untyped int constant) as InputMethodQuery value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:1310:46: cannot use 4294967295 (untyped int constant) as InputMethodQuery value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:1338:46: cannot use 4294901760 (untyped int constant) as InputMethodHint value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:1564:34: cannot use 4294967295 (untyped int constant) as GestureType value in constant declaration (overflows)
/src/qt/gen_qnamespace.go:1564:34: too many errors
```

PR #115, opened 2024-12-30
Support building on 32-bit platforms

mappu:
Fixes: #3 

The only thing preventing using Miqt on 32-bit was some large integer constants. Move them to a separate `_64bit.go` file that is blocked on `386` and `arm` GOARCH.

This is not a complete list of 32-bit GOARCH but it's the only common ones. More could be added. There is no builtin build tag for "32-bit".

Add 32-bit Windows docker containers.

### DIFF
--- a/cmd/genbindings/emitgo.go
+++ b/cmd/genbindings/emitgo.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"go/format"
 	"log"
+	"math"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -643,7 +645,7 @@ func (gfs *goFileState) emitCabiToGo(assignExpr string, rt CppParameter, rvalue 
 
 }
 
-func emitGo(src *CppParsedHeader, headerName string, packageName string) (string, error) {
+func emitGo(src *CppParsedHeader, headerName string, packageName string) (string, string, error) {
 
 	ret := strings.Builder{}
 	ret.WriteString(`package ` + path.Base(packageName) + `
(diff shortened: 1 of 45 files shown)


qt6/gen_qcalendar.go
```diff
@@ -15,10 +15,6 @@ import (
 
 type QCalendar__ int
 
-const (
-	QCalendar__Unspecified QCalendar__ = -2147483648
```
rcalixte, line 19: This should fit in a 32-bit integer I thought.

qt6/gen_qinputdevice_64bit.go
```diff
@@ -0,0 +1,8 @@
+//go:build !386 && !arm
+// +build !386,!arm
+
+package qt6
+
+const QInputDevice__AllDevices QInputDevice__DeviceType = 2147483647
+const QInputDevice__All QInputDevice__Capability = 2147483647
```
rcalixte, line 7: These should also fit inside a 32-bit integer as well.

cmd/genbindings/emitgo.go
```diff
@@ -716,13 +720,41 @@ import "C"
...
+				if e.UnderlyingType.IntType() {
+					// Int-type enums need special handling in case of 64-bit
+					// overflow, that would prevent using miqt on 32-bit platforms
+
+					entryValueI64, err := strconv.ParseInt(ee.EntryValue, 10, 64)
+					if err != nil {
+						panic("Enum " + ee.EntryName + " has non-parseable integer value")
+					}
+
+					if entryValueI64 >= math.MaxInt32 || entryValueI64 <= math.MinInt32 {
```
rcalixte, line 738: This can technically drop the equality portions.

```suggestion
					if entryValueI64 > math.MaxInt32 || entryValueI64 < math.MinInt32 {
```

mappu:
@rcalixte Thanks for the review, I've applied those changes, it does make it simpler. :+1:

PR #118, opened 2025-01-03
clang2il.go: Check for desugared types (support clang-19+)

### LINKED ISSUE #116: `genbindings` doesn't work with `clang-19`
Just tried the bindings generator on a more recent distro than bookworm, where clang 19 is available - the json output has changed and the way constructor types are specified has changed as well:

```json
			{
				"id": "0x564dc729da10",
				"inner": [
					{
						"id": "0x564dc729d8f0",
						"kind": "ParmVarDecl",
						"loc": {
							"col": 26,
							"offset": 13690,
							"tokLen": 4
						},
						"name": "type",
						"range": {
							"begin": {
								"col": 21,
								"offset": 13685,
								"tokLen": 4
							},
							"end": {
								"col": 26,
								"offset": 13690,
								"tokLen": 4
							}
						},
						"type": {
							"desugaredQualType": "QEvent::Type",
							"qualType": "Type"
						}
					}
				],
				"kind": "CXXConstructorDecl",
				"loc": {
					"col": 14,
					"line": 297,
					"offset": 13678,
					"tokLen": 6
				},
				"mangledName": "_ZN6QEventC1ENS_4TypeE",
				"name": "QEvent",
				"range": {
					"begin": {
						"col": 5,
						"offset": 13669,
						"tokLen": 8
					},
					"end": {
						"col": 30,
						"offset": 13694,
						"tokLen": 1
					}
				},
				"type": {
					"qualType": "void (Type)"
				}
			},
```

This is the `QEvent` constructor taking a `QEvent::Type` - in `clang-14`, the constructor type is `void (QEvent::Type)` while in clang-19 it's `void (Type)` - the consequence is that `Type` is no longer found as an "enum" leading to lots of downstream problems.

To support `clang-19`, one would have to dig into `inner` and fetch the type from `desugaredQualType` instead of parsing the prototype - opening this issue mostly to document the problem, even if this is not a supported way of running the generator _today_.

rcalixte:
* Closes #116


rcalixte:
I'm working on attaching the full diff with a clean cache now. I'll push it up shortly.

rcalixte:
Hmm, looking at the diff closer, it seems to remove a few functions. Should those be removed or should we look for re-adding them back in?

mappu:
The missing functions do not look very important, we could live without them if it makes clang-19 work.

I think it is fixable -

Here's my (unconfirmed) theory:

Taking qt6 qbrush.h QGradient::Stops as an example, the typedef has changed from

```json
		"type": {
			"desugaredQualType": "QList\u003cstd::pair\u003cdouble, QColor\u003e\u003e",
			"qualType": "QList\u003cQGradientStop\u003e"
		}
```

With the `desugaredQualType`, the QGradientStop gets expanded immediately into std::pair, and std::pair is blocklisted :scream: in config-allowlist.go. By comparison, in qt5 this same typedef is present as

```json
		"type": {
			"desugaredQualType": "QVector\u003cQPair\u003cdouble, QColor\u003e\u003e",
			"qualType": "QVector\u003cQGradientStop\u003e"
		}
```

which is apparently understood better.

In that case, one possible solution would be to remove the block on std::pair. If that's not enough by itself, it may also require handling in intermediate.go QPairOf().

Miqt's comment in config-allowlist.go:318 about QGradientStop is definitely outdated :smile:

rcalixte:
> In that case, one possible solution would be to remove the block on std::pair. If that's not enough by itself, it may also require handling in intermediate.go QPairOf().

Good call. We now have support (however infrequent 😅) for `std::pair` now. It's on my list for after I release what I'm working on to dig into other `std::` types that we can bind without too much fuss but this is a good start to tearing down some of those walls I think. Thanks again!

rcalixte:
> The remaining part of the diff converts many int types (uint64, intptr, ptrdiff) to C types (unsigned long long, etc).

We could always revert these back in emitcabi.go right? I don't mind doing that or reworking the logic for the desugared types. We could skip certain desugared types or just redo the logic entirely.

arnetheduck:
> We could always revert these back in emitcabi.go right? 

Technically the c sources become a little bit more portable if they retain `intptr_t` etc, ie they better capture the original qt intent - you can't get these back from `int`, `long` etc since they are different depending on os/cpu/abi/gcc flags sometimes/etc - above all, if the generated C files are to be used from C, this could be a problem.

rcalixte:
> Technically the c sources become a little bit more portable if they retain `intptr_t` etc, ie they better capture the original qt intent - you can't get these back from `int`, `long` etc since they are different depending on os/cpu/abi/gcc flags sometimes/etc - above all, if the generated C files are to be used from C, this could be a problem.

Maybe it's a hack (just wait until you see what I've done 😆), but what if we exclude those types from the desugaring? I'm thinking a conditional for `_t` types and/or `intptr` types or something along that course.

arnetheduck:
I guess one can grab a list from https://pubs.opengroup.org/onlinepubs/009604599/basedefs/stdint.h.html and it would hopefully cover most of it? always a mess with c, finding the right level of de-macro-and-typedef-ification ;) it's not an unreasonable starting point at least.

rcalixte:
> it's not an unreasonable starting point at least.

I ended up grabbing the Qt typedefs from https://doc.qt.io/qt-6/qttypes.html too. The diff didn't change much without it.

arnetheduck:
:+1: 

https://pubs.opengroup.org/onlinepubs/7908799/xsh/stddef.h.html is another source, pre-c99 so probably also good to cover

cmd/genbindings/clang2il.go
```diff
@@ -165,6 +165,29 @@ nextTopLevel:
 	return &ret, nil // done
 }
 
+// shouldPreferQualType returns true if we should use the qualType instead of
+// the desugared type based on certain type patterns
+func shouldPreferQualType(qualType string) bool {
+	if strings.Contains(qualType, "intptr") || strings.Contains(qualType, "_t") ||
+		strings.HasPrefix(qualType, "uint") || strings.Contains(qualType, "ushort") ||
+		strings.Contains(qualType, "quint") || strings.Contains(qualType, "qint") ||
+		strings.Contains(qualType, "qptrdiff") || strings.HasPrefix(qualType, "qsize") {
```
rcalixte, line 174: @arnetheduck Thoughts? 😅 

This diff is still showing more changes than it should so I'm looking into why that is.

rcalixte: This is why. 😔 

```
DESUGARED: long long
QUALTYPE: QIntegerForSize<8>::Signed
DESUGARED: unsigned long long
QUALTYPE: QIntegerForSize<8>::Unsigned
DESUGARED: unsigned long long
QUALTYPE: QIntegerForSizeof<void *>::Unsigned
DESUGARED: long long
QUALTYPE: QIntegerForSizeof<void *>::Signed
```

cmd/genbindings/clang2il.go
```diff
@@ -512,8 +553,16 @@ func processEnum(node map[string]interface{}, addNamePrefix string) (CppEnum, er
...
-		if nodequal, ok := nodefut["qualType"].(string); ok {
-			ret.UnderlyingType = parseSingleTypeString(nodequal)
+		var desugared, qualType string
+		if d, ok := nodefut["desugaredQualType"].(string); ok {
+			desugared = d
+		}
+		if q, ok := nodefut["qualType"].(string); ok {
+			qualType = q
+		}
+		qualType = getPreferredType(desugared, qualType)
```
mappu, line 563: I think this function could take the node `interface{}` as a parameter, that would simplify all the call sites.

cmd/genbindings/clang2il.go
mappu: There is a list of int types already in intermediate.go IntType() that may be useful, it could be put in a common function.

There are some places in genbindings where we do need to accurately know the type before typedefs are applied, in order to (e.g.) correctly hit a precice overload of a function. But if clang itself is considering these types equivalent, then it shouldn't cause any problems.

cmd/genbindings/clang2il.go
rcalixte: Agreed. Unreasonably sloppy on my part. This should be ready for review now.

qt-restricted-extras/qscintilla6/gen_qsciscintilla.cpp
```diff
@@ -3433,15 +3433,15 @@ class MiqtVirtualQsciScintilla final : public QsciScintilla {
 		struct miqt_string sigval1 = eventType_ms;
 		void* sigval2 = message;
 		qintptr* result_ret = result;
-		intptr_t* sigval3 = (intptr_t*)(result_ret);
```
arnetheduck, line 3436: I think ideally these `uintptr_t` and `intptr_t` would remain as-is

rcalixte: I saw that on the review too. The methods that were being removed still need to be addressed as well. I'll see if I can get that sorted now.

rcalixte:
This managed to get both hackier and cleaner at the same time. 😔

mappu:
This is looking great. In the end, there was no diff to the Qt types at all, that's actually fantastic.

I'll try it with clang-19 on my PC today, and if it all works cleanly, let's merge it.

mappu:
I tried this on commit 24cf11dc28a3ad800da34450548f3efa402285f5. I installed clang-19 from apt, and cleared genbindings' cache directory with all existing json files. But there's a panic,

```bash

# Check my existing Clang version
$ clang --version
Debian clang version 14.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

# Install Clang 19 side-by-side
$ apt install clang-19
[...]

$ clang-19 --version
Debian clang version 19.1.4 (1~deb12u1)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-19/bin

# Clear genbindings' existing cache
$ tar caf cachedir-clang-14.tar.xz cachedir/ # Saving it for later, since it takes a while to regenerate fully
$ rm -r cachedir
$ mkdir cachedir

# Genbindings has a flag argument to choose the path to the clang compiler
$ ./genbindings -clang clang-19

[...]
Processing "/usr/include/x86_64-linux-gnu/qt5/QtCore/qabstractanimation.h"...
Blocking method "qt_metacall"([Param(QMetaObject::Call param1) Param(int param2) Param(void** param3)]): Type declaration is too complex to parse
Blocking method "qt_metacall"([Param(QMetaObject::Call param1) Param(int param2) Param(void** param3)]): Type declaration is too complex to parse
panic: emitgo::emitCabiToGo missing type handler for parameter Param(State )

goroutine 1 [running]:
main.(*goFileState).emitCabiToGo(0xc0000fa358, {0x5b8733, 0x7}, {{0x0, 0x0}, {0xc000432371, 0x5}, 0x0, 0x0, 0x0, ...}, ...)
        /home/.../miqt/cmd/genbindings/emitgo.go:643 +0x315e
main.emitGo(0xc000198300, {0xc0001cc029?, 0x602ae8?}, {0x5b7916, 0x2})
        /home/.../miqt/cmd/genbindings/emitgo.go:907 +0x3f4e
main.generate({0x5b7916, 0x2}, {0xc000191ee0?, 0x3?, 0x40e867?}, 0x40?, {0x7ffc6bf03f9d, 0x8}, {0xc00018a000, 0xd4}, ...)
        /home/.../miqt/cmd/genbindings/main.go:225 +0xeb0
main.ProcessLibraries({0x7ffc6bf03f9d, 0x8}, {0x5b8020, 0x6}, {0x7ffc6bf03f7e, 0x17})
        /home/.../miqt/cmd/genbindings/config-libraries.go:15 +0x1b3
main.main()
        /home/.../miqt/cmd/genbindings/main.go:340 +0x14e
```

Something about the typedef resolution is damaged, it's reading `State` as a type instead of having resolved its typedef.

rcalixte:
> Something about the typedef resolution is damaged, it's reading `State` as a type instead of having resolved its typedef.

I played with this again today. As early as Clang 16, typedefs for enums aren't properly being resolved. I'm still going to prioritize getting this fixed but it's a draft pull request for now until it is ready for review/merge.

rcalixte:
I reverted this back to a draft. It works well with Qt 6.4 as the base. Qt 6.8 is still very messy.

rcalixte:
The resulting diff is much larger now due to the changes in `emitcabi.go` to support the coming QSpan container types. Both QSpan and QList have `.size()` methods but only QList has `.length()` so I switched that since they are the same operation and it trivializes supporting QSpan in 6.8. I've patched what I can based on the code drift between the upstream and downstream libraries that has occurred over time but I haven't yet done a full run on the 6.8 base with this branch. I plan on spending some time doing code cleanup with the downstream projects and whatnot after releasing the 6.8 base so if anything needs to be patched and hasn't been handled, I can pass those on. The biggest thing that comes to mind is handling for QAnyStringView.

The main reason I haven't run this branch on 6.8 yet is related to #245. For the two downstream libraries, I have removed the iterator types from the 6.8 base. After spending much time with them, they remained an obstacle and as you can see here, they required much manual patching as compared to the remainder of the type resolution. If the iterator types remain in the library, the type resolution on the 6.8 base will need to be checked and possibly patched if someone wants to approach it from the standpoint of retention in the library.

rcalixte:
I had a challenge with what I thought would be a cosmetic change on some container type resolutions. I left them alone because they were the same result but that was on Qt 6. With Qt 5, that isn't going to be an option it seems. Worse yet, this (ugly) patch to fix Qt 5 only works on Debian 12. On Debian 13 or anything with Clang 15+, we're right back where we started. I don't know if it makes sense to split out the two, especially with the issues with the libraries like QTermWidget that don't appear to exist in more than one version simultaneously. Maybe a Qt 5 base pinned to Debian 12 where there are still auxiliary libraries that are compatible with Qt 5? This was why I versioned the repositories for each downstream project.

This end result here leaves me with mixed feelings. First the iterator types and now Qt 5. Let me know how you want to proceed. I should have the Qt 6.8 base pushed to the downstreams by the end of this weekend if you want to take a look at anything there.

rcalixte:
Starting from the latest changes here, I was able to get a clean build of the helloworld example on a Qt 6.8 base with a few patches that are not part of this pull request:

* Removing the aforementioned iterator types
* Patching gen_qflags.cpp with `#include <new>` to support the usage of `std::nothrow`
* Patching gen_qpermissions.* to remove signals that invoke invalid `::connect` methods (downstream, I just removed these)
* Deleting binding.go since `QSysInfo::Endian` is now properly generated resulting in a duplicate declaration (downstream, quite a few of the manual typedefs/workarounds are cleaned up)

Other examples that don't require any modification for the Qt 6.8 changes also work just fine. There is a curious regression with private signals that I haven't dug into yet but this didn't require any patching downstream so I wonder if something else is leading to the block.

The CI is failing. I can try to patch the fix but I can't replicate it outside of CI so I wonder if it needs to be patched at all.

rcalixte:
Last touch up I promise. 😅

rcalixte:
> There is a curious regression with private signals that I haven't dug into yet but this didn't require any patching downstream so I wonder if something else is leading to the block.

There must be some over-zealous matching done somewhere else that escapes me at the moment. The same definitions and parsing existing in clang2il.go and config-allowlist.go as the downstreams and worse yet, this only appears when run against Qt 6.8 not Qt 6.4. When I comment out the `QPrivate` and `QPrivateSignal` conditionals in config-allowlist.go, the private methods incorrectly come through as regular methods (meaning direct invocations in addition to the callbacks). The block otherwise:

```text
Blocking method "timeout"([Param(QPrivateSignal param1)]): Type declaration is too complex to parse
```

To add to the curiosity, commenting out either one of the checks and leaving the other has no effect.

PR #123, opened 2025-01-07
Small fixes

arnetheduck:
* `stdlib` is not used / needed
* `miqt_strdup` doesn't exist
* consistently use nameprefix for `_ret`


mappu:
LGTM :ship:

The miqt_strdup and stdlib were leftover from a previous (v0.5.0-era) way of marshaling strings and are no longer required.

PR #124, opened 2025-01-07
Avoid dependency on exported go header

arnetheduck:
The expected type of the callback is already known from the AST - this change reduces dependency on cgo specifics and makes the generated C ABI entirely cgo-independent - in particular, there is no need to include `_cgo_export.h` any more.


mappu:
This change makes sense to me. The cgo_export was only needed for those definitions, but if we write the prototype ourselves, the result is more portable and we avoid the weird _Bool hack.

Thanks! LGTM :ship:

PR #126, opened 2025-01-09
Add `Makefile` for rebuilding bindings

arnetheduck:
Having a makefile is a nice little trick to re-run the binding generation code without having to copy-paste and/or remember the right docker incantations :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ examples/libraries/restricted-extras-qscintilla/restricted-extras-qscintilla
 # android temporary build files
 android-build
 deployment-settings.json
+
+*.docker-buildstamp
+compile_flags.txt
--- a/.github/workflows/miqt.yml
+++ b/.github/workflows/miqt.yml
@@ -13,10 +13,7 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-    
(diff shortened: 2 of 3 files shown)


arnetheduck:
Oops - forgot that docker mounts are owned by root by default - push a version that sets the user instead

.gitignore
```diff
@@ -46,3 +46,6 @@ examples/libraries/restricted-extras-qscintilla/restricted-extras-qscintilla
 # android temporary build files
 android-build
 deployment-settings.json
+
+*.docker-buildstamp
+compile_flags.txt
```
mappu, line 51: What is compile_flags.txt?

mappu:
Thanks, this looks great. I have a similar helper script that sets up local disk paths for my dev environment, but having a good one in the repo already will be nice.

.gitignore
arnetheduck: ah, that's something [`clangd` uses](https://clangd.llvm.org/installation#compile_flagstxt) to provide (more) accurate intellisense - can remove or leave a comment, whichever you prefer.

mappu: If it's for something well-known like clangd LSP, that's fine as-is. :ok_hand:

Issue #125, opened 2025-01-09
gen_qhashfunctions.cpp compilation error on Windows (MSYS2 UCRT64, Qt6 6.7.2): "QHashSeed has no member named 'operator long unsigned int'"


Ye1v:
When building the project mappu/miqt on Windows using MSYS2 with the UCRT64 environment and mingw-w64-ucrt-x86_64-qt6-base-6.7.2-2-any.pkg.tar.zst, the compilation fails with the following error:
```
angelscript

# github.com/mappu/miqt/qt6
gen_qhashfunctions.cpp: In function 'size_t QHashSeed_ToUnsignedLong(const QHashSeed*)':
gen_qhashfunctions.cpp:19:40: error: 'const class QHashSeed' has no member named 'operator long unsigned int'
   19 |         return self->operator unsigned long();
      |                                        ^~~~~

```

mappu:
I can confirm the issue. Outside of MSYS2, it also affected mingw-w64 packages on Fedora 41.

```
[root@8383a800ee74 helloworld6]# go build -ldflags '-s -w'# github.com/mappu/miqt/qt6
gen_qhashfunctions.cpp: In function 'size_t QHashSeed_ToUnsignedLong(const QHashSeed*)':
gen_qhashfunctions.cpp:22:40: error: 'const class QHashSeed' has no member named 'operator long unsigned int'
   22 |         return self->operator unsigned long();
      |                                        ^~~~

```

Actually the package QHashSeed never had a `operator unsigned long()`, it had only a `operator size_t()`. - Qt source code https://github.com/qt/qtbase/blob/v6.8.0/src/corelib/tools/qhashfunctions.h#L40

The bindings were generated using Linux x86_64. On that platform, `size_t` is an alias for `unsigned long`, but on mingw-w64-x86_64 it's an alias for different `unsigned int`.

Probably, the best solution is that the typedef should not be resolved, the code should say here `return self->operator size_t();`.

mappu:
It should now be fixed on `master`. Also, the Github CI will prevent it from regressing in the future. :muscle:

PR #128, opened 2025-01-10
Fix Compilation issues/125 on Windows with UCRT64 Environment 

Ye1v:
This pull request addresses compilation issues encountered when building the mappu/miqt project on Windows using the MSYS2 environment with UCRT64

### DIFF
--- a/qt6/gen_qhashfunctions.cpp
+++ b/qt6/gen_qhashfunctions.cpp
@@ -16,7 +16,7 @@ QHashSeed* QHashSeed_new2(size_t d) {
 }
 
 size_t QHashSeed_ToUnsignedLong(const QHashSeed* self) {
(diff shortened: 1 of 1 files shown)


mappu:
Fixes: #125 

Hi @Ye1v , thank you for the PR!

Sorry, but you can't change the gen_ files directly, because they will be overwritten again by the code generator. The change must be made inside the cmd/genbindings program.

One solution is: this function can be blocklisted from generation in config-allowlist.go (it seems uncommon to ever call this function)

Another solution is: the root cause of Windows incompatibility can be addressed in some other way

mappu:
Closed in favor of #131. Thanks again for the PR and for doing all the research work for this issue. In the end, the fix was similar.

PR #132, opened 2025-01-11
Move `go` name mangling to `emitgo`

arnetheduck:
This change helps keep rules for each language separate by moving `go` rules to `emitgo` while the C bindings stay closer to the original Qt naming (that already is mostly C-safe).

Although it doesn't practically matter for go, it makes it slightly easier to reuse the generated code in other languages which have different keywords and naming conventions.

The cabi generator also gains a few helpers to help keep names consistent across files which hopefully aids reading the generator code
- it did for me at least;)

The rule that converts under_score to CamelCase is left for another day since moving it turns out to be more invasive due to name collision handling - when underscores are kept, there are fewer name conflicts which ends up causing name changes in  the public go api when done naively.


cmd/genbindings/emitgo.go
```diff
@@ -640,7 +657,7 @@ func (gfs *goFileState) emitCabiToGo(assignExpr string, rt CppParameter, rvalue
 		return assignExpr + "(" + rt.RenderTypeGo(gfs) + ")(" + rvalue + ")\n"
 
 	} else {
-		panic(fmt.Sprintf("emitgo::emitCabiToGo missing type handler for parameter %+v", rt))
+		return "int /* TODO  */" //panic(fmt.Sprintf("emitgo::emitCabiToGo missing type handler for parameter %+v", rt))
```
mappu, line 660: :eyes:

ddkwork:
In the meantime, many IFs suggest switching to Switch

cmd/genbindings/emitgo.go
arnetheduck: oops, [31698df](https://github.com/mappu/miqt/pull/132/commits/31698df7838de542e32f5fabbd4cc556de649d3f)

cmd/genbindings/emitcabi.go
```diff
@@ -14,10 +14,58 @@ func cppComment(s string) string {
...
+	parmName := p.ParameterName
+	if cReservedWord(parmName) {
+		parmName += "Val"
+	}
+
+	return parmName
+}
+
 func cabiCallbackName(c CppClass, m CppMethod) string {
 	return "miqt_exec_callback_" + cabiClassName(c.ClassName) + "_" + m.SafeMethodName()
```
arnetheduck, line 38: could update this one to the below naming pattern, potentially.

arnetheduck:
> many IFs suggest switching to Switch

`+332,433 −332,376` feels like enough changes for a single PR :stuck_out_tongue:

mappu:
The helpers for virtbase, delete, etc are actually pretty useful, i like them.

I can't comment on the usefulness of this for other language bindings, but i can see it being important especially if the set of reserved words needs to change a lot. Although it doesn't affect Go, if a small change like this can help keep all other miqt versions closer together, that's a good thing.

I have pushed two commits to your branch, to rebase onto master (one last time...) and to split the genbindings + rebuild changes into two separate commits. Once this passes the pipeline, let's merge it :ship:

mappu:
Whoops, merging this broke some examples (modelview_color6)

arnetheduck:
Great! 

> Whoops, merging this broke some examples (modelview_color6)

Could these be compiled as part of CI, as a sanity check?

PR #135, opened 2025-01-15
Log blocked methods

arnetheduck:
Blocked methods are the source of many gaps in the binding - in particular, they block qt_metacall/activate/etc that is necessary for implementing meta-object support.

This change makes them visible in logs and also removes log timestamps so that logs from two runs easily can be diffed.


mappu:
Nice, this makes sense to me.

Issue #133, opened 2025-01-11
Remove isSubclass from Go structs


mappu:
> `isSubClass` looks redundant - since destructors are virtual (are they, all?), they should not need the extra `dynamic_cast` - potentially, an ownership tracker could be introduced here

_Originally posted by @arnetheduck in https://github.com/mappu/miqt/issues/21#issuecomment-2573223116_

---

The isSubClass field is used to prevent trying to call QWidget_override_virtual_ functions on a QWidget that was created externally to Miqt (e.g. if you create a QTreeWidget, the default QHeaderView is created by Qt and cannot be subclassed).

We are carrying an extra bool to check this Go-side. But, the override_virtual implementation could check the result of dynamic_cast<> for a nullptr response, and return false back to Go space. Then Go can have the same safety check without needing an extra struct field.

If there is a panic, notifying the developer costs an extra ABI roundtrip, but you should never see this panic in a working application, only in early development.

arnetheduck:
For subclassing, an alternative that maybe could work is something like the following, assuming X is the Qt type:

```c++
struct MitqVtableX {
  // include both this_ and vtbl so the host language can know / cast these without having to store them in a closure
  void (*override1)(X* this_, MitqVtableX* vtbl, ...);
  void (*override2)(X* this_, MitqVtableX* vtbl, ...);
  // arbitrary data that the host language can use for its own bookkeeping (optional, 
  // since you can simply `malloc(sizeof(MiqtVtableX) + sizeof(void*))` in the host language 
  // which has the same effect)
  void *extra_data; 
};

struct MiqtVirtualX : X {
  MiqtVirtualX(const MitqVtableX* vtbl, void (*destructor)(MiqtVtable* vt)) : vtbl(make_unique(vtbl, destructor)) {}

  std::unique_ptr<MiqtVtableX, ...> vtbl;
  
  void override1(...) override { vtbl->override1(this, ...); }
};

extern "C" X* X_new_override(const MitqVtableX* vtbl, ... destructor, ...) { ...}

```

This would provide a few things:

* the vtable for overriding things is set once at construction which means there's no c api that needs to check for panics
* since qt _generally_ manages lifetimes via the parent/child relationship, this provides a hook for the host language to do its own cleanup of the vtable when the subclassed object is deleted, similar to a virtual destructor

[...]

Issue #137, opened 2025-01-16
Missing overrides for virtual functions causes dynamic cast failure


arnetheduck:
* [QAbstractListModel](https://doc.qt.io/qt-6/qabstractlistmodel.html) inherits from `QAbstractItemModel` 
* creating a `QAbstractListModel` instantiates a [`MiqtVirtualQAbstractListModel`](https://github.com/mappu/miqt/blob/3d864cdb7b1ec79042edf310829d0c5f5040d47f/qt/gen_qabstractitemmodel.cpp#L3652)
* `QAbstractItemModel` contains `https://doc.qt.io/qt-6/qabstractitemmodel.html#columnCount` as a pure virtual function
* the generated bindings for `QAbstractListModel` _do not_ contain an override function for `columnCount`
* Instead, `QAbstractListModel_override_virtual_columnCount` is calledvia inheritance which fails its `dynamic_cast<MiqtVirtualQAbstractItemModel>` (because we created `MiqtVirtualQAbstractListModel`).

mappu:
There is an override function for `QAbstractListModel::columnCount` -  https://github.com/mappu/miqt/blob/master/qt6/gen_qabstractitemmodel.cpp#L547

There is also a working example for this Qt class, in the examples folder. Can you repro this on miqt master?

arnetheduck:
What you linked actually is part of `MiqtVirtualQAbstractItemModel` and as a consequence calls [miqt_exec_callback_QAbstractItemModel_ColumnCount](https://github.com/mappu/miqt/blob/501227d877d5d3867bbff7b9dcdfe25b12281f70/qt6/gen_qabstractitemmodel.cpp#L556) - notice `Item` vs `List`.

arnetheduck:
Ok, I think I understand this one now - `columnCount` has a private override in `QAbstractListModel` - so when you're using C++ it's "off limits" if you have a `QAbstractListModel` pointer. 

But there is nothing in the exported API that prevents it from being called through the inheritance structure, and this is where the failure happens.

I *think* this happens in the go bindings as well (don't know go enough), but to fix and "block" onColumnCount for the QAbstractListModel, the IL would have to retain information about private method overrides.

Normally, users wouldn't try to override this way, it doesn't make much sense, but this happened in a generic override section of nimqml that ideally would have seen a compile-time error.

mappu:
[...]

Issue #21, opened 2024-08-25
Reuse the C ABI bindings to generate Qt bindings for other non-Go languages


mappu:

arnetheduck:
Played around a bit with this idea in https://github.com/arnetheduck/nim-miqt-poc/pull/1 - works fine - miqt codebase makes it very easy to adapt/copypaste around, great stuff ;)

Here's a laundry list of questions and random thoughts that came to my mind while doing the POC, in no particular order:

* Instead of converting `miqt_string/map/array` in the C wrapper, it _maybe_ makes sense to export these as-is from qt by wrapping `QString` et al in some shape or form - generics make this a bit tricky and perhaps a middle ground would have to be found here - however, for strings and bytes at least, the Nim wrappings at least would be better off exposing `QString` to the "user" of the library (much like QT itself allows a choice between `std::string` and `QString` with converters
* the clang->il conversion could probably stay a little bit closer to C as well (naming, etc)
* some of the cgo:isms aren't really needed, like `_cgo_export.h` - all the information is already there in the clang AST
* `isSubClass` looks redundant - since destructors are virtual (are they, all?), they should not need the extra `dynamic_cast` - potentially, an ownership tracker could 
* `virtual` inheritance looks unnecessary - use `QAbstractButtton` as example, nothing will ever inherit from `MiqtQAbstractButton` so it doesn't look like it needs to virtually inherit from `QAbstractButton` - `MitqQPushButton` for example never touches `MiqtQAbstractButton`.

Regarding non-go languages, one candidate is `c` itself ;) Looking at the generated header files, they actually look very easy to use from a native C project - seen as such:

* the `clang` step could trivially be done using a Makefile
* the `c++` -> `c` step could potentially be done using Qt/C++/CMake itself to reduce the language tax for any one target language - c++ is needed anyway - though this tax is somewhat mitigated by this being a "generator"-only concern - users of wrappers don't need the extra foreign language.
* the upstream [`shiboken6`](https://doc.qt.io/qtforpython-6/shiboken6/shibokengenerator.html) project works more or less the same way - either its generator or the [xml files](https://github.com/pyside/pyside-setup/tree/dev/sources/pyside6/PySide6) it outputs could maybe serve as a predigested source of binding information - there are both benefits and risks here.
 
[...]

PR #138, opened 2025-01-18
Remove isSubclass

mappu:
Fixes: #133

Updates: #137

Updates: #21 

Currently, miqt uses the `isSubclass` variable to track whether a cgo pointer to a Qt object is for a Miqt subclass or not. This affects (A) destructor behaviour and (B) method override behaviour inside the C ABI.

- Change miqt subclasses to be final and non-virtual (Authored-By: @arnetheduck)
- Change miqt subclass destructors to ensure we always use a virtual destructor, removing reliance on isSubclass sentinel flag (Authored-By: @arnetheduck)
- Change the override functions to use the return value of `dynamic_cast<>` as a sentinel

Test that override protection is still working:

```golang
	list := qt.NewQTreeWidget2()

	header := list.Header()

	header.OnActionEvent(func(super func(event *qt.QActionEvent), event *qt.QActionEvent) {
		panic("Shouldn't be able to reach this part")
	})
```

Result:

```
panic: miqt: can only override virtual methods for directly constructed types
```

With this, it was possible to remove `isSubclass` from the language projection side.

PTAL @arnetheduck - I cherry-picked your changes from 339a2c9569eb4 in your fork, using the original naming functions for now,


arnetheduck:
lgtm.

> using the original naming functions for now,

lmk if you want to keep #132 around - it'll either need rebasing or closing.

PR #153, opened 2025-01-28
`const` in callbacks

arnetheduck:
Since callback signatures are now generated in C++, might as well go ahead and retain `const` and the pointer type, removing a `const_cast` in the process.


mappu:
Thank you - the `void*` was a holdover from when the dynamic_cast stuff was a bit more dodgy. It's great to have the real type parameter present there.

I was worried about the `const` comment about CGO compatibility, but i tried some small examples and everything works fine in my testing.

:ship:

mappu:
It looks like this only changed the cpp, not the header, so the function prototype differs from the implementation. That's probably why there were no issues with the Go side.

Issue #151, opened 2025-01-26
macos x64 qt5.15 dtls issue


qizhanchan:
os: macos x64 14.6

i trie to build qt-multimedia example
```
export PKG_CONFIG_PATH="/opt/homebrew/opt/qt@5/lib/pkgconfig"
export CGO_CXXFLAGS=-std=c++17
go build -ldflags '-s -w'
```

got the output
```
# github.com/mappu/miqt/qt/network
In file included from gen_qdtls.cpp:3:
In file included from /usr/local/Cellar/qt@5/5.15.16/lib/QtNetwork.framework/Headers/QDtls:1:
/usr/local/Cellar/qt@5/5.15.16/lib/QtNetwork.framework/Headers/qdtls.h:52:1: error: static assertion failed due to requirement 'bool(-1 == 1)': Required feature dtls for file /usr/local/Cellar/qt@5/5.15.16/lib/QtNetwork.framework/Headers/qdtls.h not available.
/usr/local/Cellar/qt@5/5.15.16/lib/QtCore.framework/Headers/qglobal.h:87:36: note: expanded from macro 'QT_REQUIRE_CONFIG'
/usr/local/Cellar/qt@5/5.15.16/lib/QtCore.framework/Headers/qglobal.h:121:49: note: expanded from macro 'Q_STATIC_ASSERT_X'
```

it seems that brew qt5 disable the dtls feature by default, refer [https://github.com/Homebrew/brew/issues/16529](https://github.com/Homebrew/brew/issues/16529)

Since it's hard to build qt from source code, any advice to pass the dtls check?

mappu:
QNetwork is only bound in miqt because QWebEngine has it as a needed dependency, and miqt tries to bind against every function. Some of them are really not useful, it would be more logical to use a DTLS function from Go.

I think we can change Miqt to exclude DTLS functions from QNetwork too. That should solve the issue for Brew's version of Qt.

Until then, a workaround would be to fork the Miqt repo and delete the 3 `qt/network/gen_qdtls` files, or, any file that mentions dtls. Then, it should compile correctly.

mappu:
This issue affects Qt 5 on Brew on macOS only.

It is because Brew builds Qt 5 using Apple Secure Transport instead of OpenSSL, and Apple Secure Transport does not support DTLS:
- https://github.com/Homebrew/brew/issues/16529
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/q/qt%405.rb

[...]

Issue #59, opened 2024-11-06
Add bindings for platform-specific window systems (X11, Cocoa, Win32 HWND)


mappu:
Options:
- Expose as opaque types, or
- Parse the C++ headers multiple times using each platform's compiler, and then generate build tags for each operating system

rcalixte:
I've got an interesting similar discussion opened by @fuddlesworth (rcalixte/libqt6zig#167) looking for precisely this functionality. I have some ideas for how I want to design this and ship it. I don't know if this is high on the list of priorities for either you or @arnetheduck but I figured I would share notes.

In short, I'm thinking of implementing the support via hand bindings in a package/subdirectory named `native/`. This keeps us from having to deal with the complexity of managing cross-platform external types in the generator's backend while enabling a path for whatever flexibility we might need. As much as I want to keep things in a single package/subdirectory and gate on platform to simplify things for consumers, I'm still navigating the trade-offs of that design. Many of the structs under [`QNativeInterface`](https://doc.qt.io/qt-6/qnativeinterface.html) have relatively few methods and I think too many files will create sprawl for what often amounts to static functions. There's also the caveat about source and binary compatibility but from what I've seen, we should be safe to bind to the stable portions and communicate that the remainder are unsupported until it's feasible to do so. Most of the API changes I've seen are additive.

One of the design portions I'm debating is whether or not to use the native types in the documentation and include the external headers, forcing the dependency for resolution, or whether to default to void/opaque pointers. So `#include <wayland-client.h>` to return `wl_output*` or leaving the casting to the user and returning `void*` across the board. (My other idea is that we can have the generator replace some of these types with `void*` and still generate them with the platform quirks defined.) We can probably use static casts inside the implementation so that the returned types aren't completely unsafe. To limit the guards to platform, I'm also considering only binding what's available up to the base. This still keeps the door open for some of the methods that have additional return types on newer versions of Qt while keeping the maintenance cost reasonable.

The discussion also includes methods that call into or return these types from other classes:

- [`QOpenGLContext::nativeInterface<QNativeInterface::QEGLContext>()`](https://doc.qt.io/qt-6/qopenglcontext.html#nativeInterface)
- [`QGuiApplication::nativeInterface<QNativeInterface::QWaylandApplication>()`](https://doc.qt.io/qt-6/qguiapplication.html#nativeInterface)

My thoughts are to roll these in, consistent with whatever approach is used above.

[...]

PR #159, opened 2025-01-31
Do not bind DTLS for Qt 5 QNetwork causing issues on macOS

mappu:
Fixes: #151 

Qt 5 on macOS uses Apple Secure Transport by default, including in the Brew distribution. This TLS library does not support DTLS and any attempt to use `qdtls.h` throws an `#error`.

Since DTLS is not widely used, block the binding generation for qdtls.h (on all operating systems). Future work could be to make this OS-dependent (see #59).

Additional minor changes in this PR:
- Add a `make clean-cache` .PHONY target to clear the genbindings cache


mappu:
Superseded by #199

Issue #157, opened 2025-01-31
No clear way on how to use `setFormat`  from QSyntaxHighlighter CPP API


JulianH99:
I'm trying to create a very simple JSON Syntax highlighter using the examples provided here: 
- https://github.com/leixingyu/codeEditor/blob/1afd03a6e308445369306ee235f12bac0415f914/highlighter/pyHighlight.py
- https://github.com/mhassanzadeh/JsonEditor_QT/blob/master/Editor/syntaxhighlighter.cpp
- https://doc.qt.io/qt-6/qtwidgets-richtext-syntaxhighlighter-example.html. 

As you can see, I referred to the official documentation, but I cannot find how to use the `setFormat` function from the official QSyntaxHighlighter with miqt bindings. I've tried searching for a similar function in the `s.Document()` but no luck .

I don't know if I'm missing something but I'd appreciate if you could direct me or help me on how to achieve this. 
Here's my code as reference:

```go

package highlighter

import (
	qt "github.com/mappu/miqt/qt6"
)

type JsonHighlighter struct {
	formats     map[string]*qt.QTextFormat
	highlighter *qt.QSyntaxHighlighter
	document    *qt.QTextDocument
}

func format(color string) *qt.QTextFormat {
	qColor := qt.NewQColor()
	qColor.SetNamedColor(color)

	format := qt.NewQTextFormat()
	format.Foreground().SetColor(qColor)

	return format
}

func NewJsonHiglithter(parent *qt.QTextDocument) *JsonHighlighter {
	highlighter := qt.NewQSyntaxHighlighter2(parent)
	highlighter.SetDocument(parent)

	formats := map[string]*qt.QTextFormat{
		"string": format("brown"),
		"number": format("blue"),
	}

	return &JsonHighlighter{
		highlighter: highlighter,
		formats:     formats,
		document:    parent,
	}
}

func (h *JsonHighlighter) Rehighlight() {
	stringRegex := qt.NewQRegularExpression2("\"(.*?)\"")
	// numberRegex := qt.NewQRegularExpression2("\\d+")

	stringMatch := stringRegex.GlobalMatch(h.document.ToPlainText())

	for stringMatch.HasNext() {
		match := stringMatch.Next()

		h.highlighter.SetFormat( // this function does not exists, of course
			match.CapturedStart(),
			match.CapturedLength(),
			*h.formats["string"],
		)

	}
}

```

Thanks in advance, for both the help and your amazing work on bringing good qt bindings for golang

[...]

Issue #144, opened 2025-01-24
How to implement `createIndex` on QAbstractItemModel?


Ahuge:
Hi, I am just getting started using miqt and love it.

Thanks!

I am trying to implement a custom model for a QTreeView and have been following `examples/modelview_color6` to do the `On<function>` implementations.

During the implementation of `OnIndex` I need to call the protected function `QAbstractItemModel.createIndex` as they recommend in the documentation: https://doc.qt.io/qt-6/qabstractitemmodel.html#index

Is there a better way that I should be doing this?

Thanks
-Alex

mappu:
I think it will need some changes inside Miqt to be possible. It's important, so i'll work on it soon.

See also: #141

Ahuge:
Thank you very much!

mappu:
Hi @Ahuge - I have a PR ready that will add support for protected functions like `createIndex`. Can you please update Miqt to the PR branch (go get github.com/mappu/miqt@miqt-protected-methods) and see if the generated CreateIndex() function works in your case?

Ahuge:
I will dig into it but it looks like it should work!

Ahuge:
Update I created a PR in #174 that showcases this

rcalixte:
> Update I created a PR in [#174](https://github.com/mappu/miqt/pull/174) that showcases this

Does it make sense to try to simplify the example? Looking to [modelview_color6](https://github.com/mappu/miqt/blob/2cf0252/examples/modelview_color6/main.go) for inspiration on this if possible.

Issue #141, opened 2025-01-19
Make protected functions callable from virtual override classes


arnetheduck:
When overriding abstract base classes like `QAbstractItemModel`, there's a number of functions that the overriding class is supposed to call as part of the override implementation, ie see [`beginInsertRows`](https://doc.qt.io/qt-6/qabstractitemmodel.html#protected-functions) and friends.

arnetheduck:
There's a partial implementation of this feature (for c) at https://github.com/arnetheduck/nim-miqt-poc/pull/1/commits/343888c56263bcfe03bcda898ca8c950ffa4db17 done by copy-pasting the virtualbase solution that perhaps can serve as inspiration - that said, I think this would better be achieved with friend functions, ie that would probably result in less code being generated.

Issue #163, opened 2025-02-01
Implicit assignment operator missing


arnetheduck:
Types like `QModelIndex` have an implicit assignment operator that ideally should be exposed, just like the explicit ones.

mappu:
There is definitionData->copyAssign field in the clang AST.

The same place also has fields for moveAssign / copyCtor / defaultCtor / etc. Probably the implicit constructor and copy constructor might also be useful to add bindings for.

rcalixte:
If it helps, I ended up calling these `QCopy`, `QMove`, etc. to avoid namespace collisions. It's not too bad to implement since the AST calls them out as trivial. If this doesn't have an owner by the time I release the other libraries, I wouldn't mind getting it taken care of after getting Clang 19 sorted. (Clang 19 is probably the top thing on my list to get sorted after the releases but after that, everything else is fair game where I can lend a hand.)

arnetheduck:
> OperatorAssign

The important thing to maintain here is consistency between implicit and explicit versions of this function in particular.

Regarding copy constructors, they are useful for implementing meta-object "copy" support which in turn is needed for cross-thread signal emission. While useful, they can also in most cases be "emulated" with a default constructor call + assignment, moving them off the critical implementation path.

arnetheduck:
Turns out there's a simpler explanation :)

```patch
diff --git a/cmd/genbindings/clang2il.go b/cmd/genbindings/clang2il.go
index a5e1d5e1a..c077681f7 100644
--- a/cmd/genbindings/clang2il.go
+++ b/cmd/genbindings/clang2il.go
@@ -361,10 +361,12 @@ nextMethod:
                                return CppClass{}, errors.New("method has no name")
                        }
 
-                       // If this is a virtual method, we want to allow overriding it even
-                       // if it is protected
-                       // But we can only call it if it is public
-                       if visibility == VsPrivate {
+                       if isImplicit, ok := node.Fields["isImplicit"].(bool); ok && isImplicit {
+                               // implicit is implicitly public!
+                       } else if visibility == VsPrivate {
+                               // If this is a virtual method, we want to allow overriding it even
+                               // if it is protected
+                               // But we can only call it if it is public
                                ret.PrivateMethods = append(ret.PrivateMethods, methodName)
                                continue // Skip private, ALLOW protected
                        }
```

mappu:
Fixed by #263

PR #166, opened 2025-02-06
Add bindings for protected methods

mappu:
Fixes: #141 

Fixes: #144 

Fixes: #157 

Add bindings for protected methods of Qt classes.

This is an extended version of @arnetheduck 's Nim version from #141 that uses friend methods.

Caveats:
- Can't do protected static methods (e.g. `QBitmap::fromImageInPlace`) as there is no class context
- Skips over protected operators - the subclass causes problems for implicit `operator=(other&)`.
    - Future work in #163 may help with this.
- Protected methods are only bound if there are any virtual methods
    - The presence of virtual methods triggers the generation of a subclass. Generating the subclass without a virtual base class, prevents the use of `dynamic_cast<>`, which means we lose the runtime segfault protection against misuse.
    - Future work in #137 may help with this.


cmd/genbindings/emitcabi.go
```diff
@@ -600,11 +604,54 @@ func getCppZeroValue(p CppParameter) string {
...
 	} else if p.ParameterType == "bool" {
 		return "false"
+	} else if p.ParameterType == "void" {
+		return ""
 	} else {
 		return p.RenderTypeQtCpp() + "()"
 	}
 }
 
+func getCabiZeroValue(p CppParameter) string {
```
arnetheduck, line 614: can probably get away with plain `return {};` for all the below types except `void`, simplifying the generator a little.

arnetheduck:
nice with the friend functions! can probably do the same for virtualbase

mappu:
> nice with the friend functions! can probably do the same for virtualbase

I tried this, but calling the `BaseClass::something()` syntax really needs a real class context, not just protected member access. Is there a syntax for it?

[Link to online C++ playground](https://codapi.org/embed/?sandbox=cpp&code=data%3A%3Bbase64%2CpZHLSgMxFIb3eYpDBZkW7OzTYYSC4saFuBWGTBJrIJcxl2qReXeTyYxM202xWQRyLt9%2F8p%2ByhIdvojrJobNmZ4lCN0JTGRiHShjnLSeqnsViROhdjRCVxDl4eTQGflAXWikoRgji2QvrA5GwN4LBlthiGStgPM4zjKkJHqoKFqkd41jzpheboaZH%2FWaCP4tP%2Fxra%2FMCQRa6VnENPpdOdeYkzQrfE8eaU%2BTd4sTxrfreCa5YZKqq5Ua3JiWbktonbRsJ8ohWoCEwW%2FLsd5SnLEp7MF3gDhFIeDfQfHIJm3MpD3OD8B%2FdTg7qr85fSo4uL9sdLGOqnYnbQRAnaUOJ8lWirulDLiwg9QjEDighdTPMeLVtlU9V65rBany1kTFxg0u1g6y8%3D)

arnetheduck:
>  Is there a syntax for it?

Using [qualified lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup) (class member section) should do the trick: 
```c
void miqtsubclass_friend_virtualbase_bar(MiqtSubclass* m)
{
    // How to access the underlying QFoo::Bar()?
    m->QFoo::Bar(); // prints QFoo::Bar
}
```

mappu:
The main part of this feature is working, the extra scope is split out:

- #171
- #172

cmd/genbindings/emitcabi.go
```diff
@@ -1262,6 +1347,60 @@ extern "C" {
...
+				// Also need to scan parent classes (e.g. QColumnView friend
+				// functions refer to its parent QAbstractItemView::State)
+				for _, classInherit := range c.AllInheritsClassInfo() {
+					assignStmts = strings.Replace(assignStmts, classInherit.Class.ClassName+`::`, cppSubclassName(c)+`::`, -1)
+				}
+
+				//
+
+				ret.WriteString(
+					m.ReturnType.RenderTypeCabi() + " " + cabiProtectedBaseName(c, m) + "(bool* _dynamic_cast_ok, " + emitParametersCabi(m, ifv(m.IsConst, "const ", "")+"void*") + ") {\n" +
```
arnetheduck, line 1386: The safety check with dynamic_cast is not really needed here - when using qualified lookup, the call is done using static dispatch meaning that it does not actually touch the vbtable/derived type/anything like that.

Issue #147, opened 2025-01-24
Reduce binary size


mappu:
The ideas from #8 and #9 can also help to reduce binary size.

Ideas:

- [ ] Use `-ffunction-sections -fdata-sections` in CFLAGS, and `-Wl,--gc-sections` in LDFLAGS to allow stronger dead code elimination across language boundaries
    - https://github.com/golang/go/issues/38135
- [ ] Refactor marshaling code to call helper functions instead of inlining it in every function wrapper
- [ ] Ability to disable unwanted Qt classes at compile time?
    - Maybe with a build tag
    - Similar idea to `qtminimal`
- [ ] Ability to disable subclassing support for unneeded classes?

ddkwork:
It's 17-21mb after removing cgo, compilation speed and binary size are very impressive, but I'm not very familiar with cpp classes and templates. I will continue to remove cgo when I have enough experience, I tried to manually fix the c abi file generated by miqt and compiled it to dynamic dll, it was 17mb, the outer code of the go project was about 2-3mb, it added up to about 20mb, on par with flutter.

arnetheduck:
FWIW in the nim bindings, `-flto` removes almost all overhead since it ends up including only the wrappers that actually get used - this approach of course depends on the underlying C++ compiler - in a way, it's similar to sections but more powerful.

Another idea might be to follow the upstream modules more closely - ie miqt lumps together QtCore/widgets/etc which might not be necessary for all applications - recycling the upstream qt module organisation has other benefits too since the bindings will be more in tune with available documentation etc.

mappu:
I recall trying `-flto`, but the Go compiler doesn't do anything with it, so it had no effect overall. It would probably work very well with gccgo or gollvm/llgo. I'll try that, it could be a recommendation for people to make smaller release builds with a different compiler.

There is probably only limited use cases for using a language binding of Qt Core without gui/widgets. Almost everything in Core has some sort of native equivalent in Go anyway. I can maybe imagine extending Miqt to bind some custom library that uses Core types without gui/widgets, but that's hypothetical at this stage. (I'm not sure if Qml or Quick depend on Gui.)

There are also some difficult circular dependencies in the Qt headers between package types - at least QtMultimedia and QtMultimediaWidgets, etc.

mappu:
- gccgo does not yet implement go 1.18 language features
[...]

PR #172, opened 2025-02-08
Convert virtualbase wrapper functions to friend functions

mappu:
Updates: #147 

Updates: #166 

Miqt supported calling the base/super method of any virtual override function. This worked by (A) creating a `virtualbase_...` method wrapper in the subclass, and (B) also adding a C ABI function to call it.

Combine these into a single C ABI function, following the `friend` pattern of the new protected-method support. This reduces the number of functions generated in the translation unit.


qt-restricted-extras/qscintilla/gen_qsciabstractapis.cpp
```diff
@@ -534,7 +496,9 @@ bool QsciAbstractAPIs_override_virtual_event(void* self, intptr_t slot) {
 }
 
 bool QsciAbstractAPIs_virtualbase_event(void* self, QEvent* event) {
-	return ( (MiqtVirtualQsciAbstractAPIs*)(self) )->virtualbase_event(event);
+
+	return ( (MiqtVirtualQsciAbstractAPIs*)(self) )->MiqtVirtualQsciAbstractAPIs::event(event);
```
arnetheduck, line 500: oops, instead of calling the base class function, this actually calls the override version which goes on to call the language override in a nice infinite recursion ;)

This needs to be `( (MiqtVirtualQsciAbstractAPIs*)(self) )->QsciAbstractAPIs::event(event)` + argument conversion

mappu: The `c.ClassName` reference was right, but it got caught up in the `fixupProtectedReferences` logic. Probably it will need to be teased apart so that it still builds on Android.

PR #174, opened 2025-02-11
Introduce a new example using a QSortFilterProxyModel

Ahuge:
This implements a custom QAbstractItemModel and QSortFilterProxyModel

### DIFF
--- a/examples/customsortfiltermodel/main.go
+++ b/examples/customsortfiltermodel/main.go
@@ -24,7 +24,7 @@ func PopulateData(rootItem *TreeItem) {
 	cellData = append(cellData, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("Big File")))
 	cellData = append(cellData, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("10.7 MiB")))
 	item := NewTreeItem(cellData, rootItem, false)
-	err := item.SetData(1, qt.NewQVariant11("11225400"), qt.UserRole)
+	err := item.SetData(1, qt.NewQVariant6(11225400), qt.UserRole)
 	if err != nil {
 		fmt.Printf("Error setting UserData: %v\n", err)
 	}
@@ -38,7 +38,7 @@ func PopulateData(rootItem *TreeItem) {
 	cellData2 = append(cellData2, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("Small File")))
 	cellData2 = append(cellData2, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("110 Bytes")))
 	item2 := NewTreeItem(cellData2, rootItem, false)
-	err = item2.SetData(1, qt.NewQVariant11("110"), qt.UserRole)
+	err = item2.SetData(1, qt.NewQVariant6(110), qt.UserRole)
 	if err != nil {
 		fmt.Printf("Error setting UserData: %v\n", err)
 	}
@@ -52,7 +52,7 @@ func PopulateData(rootItem *TreeItem) {
 	cellData3 = append(cellData3, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("Medium File")))
 	cellData3 = append(cellData3, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("109.5 KiB")))
 	item3 := NewTreeItem(cellData3, rootItem, false)
-	err = item3.SetData(1, qt.NewQVariant11("112200"), qt.UserRole)
+	err = item3.SetData(1, qt.NewQVariant6(112200), qt.UserRole)
 	if err != nil {
 		fmt.Printf("Error setting UserData: %v\n", err)
 	}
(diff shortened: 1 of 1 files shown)


mappu:
Nice, thank you for the PR! It compiled fine for me and sorting the Size column does correctly understand the byte value of the files.

image

examples/customsortfiltermodel/tree_cell.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (cell *TreeCell) SetData(data qt.QVariant, role qt.ItemDataRole) {
+	cell.data[role] = data.ToString()
+}
+
+func (cell *TreeCell) Data(role qt.ItemDataRole) *qt.QVariant {
+	if role == qt.UserRole {
+		number, err := strconv.Atoi(cell.data[role])
+		if err != nil {
+			fmt.Printf("Error converting %s to int: %v\n", number, err)
```
mappu, line 27: When starting up the example, I got these terminal messages:

```
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
Error converting %!s(int=0) to int: strconv.Atoi: parsing "": invalid syntax
```

They repeat when attempting to sort by the Name column,

examples/customsortfiltermodel/model.go
```diff
@@ -0,0 +1,94 @@
...
+
+	model.OnRowCount(func(parent *qt.QModelIndex) int {
+		if parent.Column() > 0 {
+			return 0
+		}
+		var item *TreeItem
+		if !parent.IsValid() {
+			item = &rootItem
+		} else {
+			item = (*TreeItem)(parent.InternalPointer())
```
mappu, line 80: There are some extra rules when passing pointers between Go and C. One of the rules is: If there is no Go pointer to the TreeItem, only a C++ pointer, then Go might GC the variable, and this could become a use-after-free.

Is there any other place this TreeItem is reachable other than from Qt (C++) memory? If not, a workaround is to either (A) ensure that the TreeItems are stored in some long-lived application struct or global variable; or (B) call runtime.Keepalive() at some later point.

examples/customsortfiltermodel/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return treeView
+}
+
+func PopulateData(rootItem *TreeItem) {
+	cellData := make([]*TreeCell, 0)
+	cellData = append(cellData, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("Big File")))
+	cellData = append(cellData, NewTreeCell(qt.DisplayRole, *qt.NewQVariant11("10.7 MiB")))
+	item := NewTreeItem(cellData, rootItem, false)
+	err := item.SetData(1, qt.NewQVariant11("11225400"), qt.UserRole)
```
mappu, line 27: The sort size is stored in a QVariant as a string. Then later, we parse it. It might be simpler to use the QVariant constructor that stores an int directly,

examples/customsortfiltermodel/main.go
```diff
@@ -0,0 +1,93 @@
...
+		leftData := sortModel.SourceModel().Data(source_left, sortModel.SortRole())
+		rightData := sortModel.SourceModel().Data(source_right, sortModel.SortRole())
+		return leftData.ToInt() < rightData.ToInt()
+	})
+
+	sortModel.SetSourceModel(model)
+	treeView.SetModel(sortModel.QAbstractItemModel)
+
+	dialogLayout.AddWidget3(treeView.QWidget, 0, qt.AlignCenter)
+	return dialog
```
mappu, line 83: In the default view, the Size column is initially truncated, so I went to resize the window larger. But something about the `QSizePolicy` is meaning the treeview will not grow when the window is resized (although it does shrink). I think this could be fixed in a 1 line change, but I haven't checked exactly.

image

image

examples/customsortfiltermodel/tree_cell.go
Ahuge: Ah thanks I will fix that!

examples/customsortfiltermodel/main.go
Ahuge: Good call

examples/customsortfiltermodel/model.go
mappu: Or (C) Wrap the thing in a cgo.Handle before putting it in Qt space, and unwrap it when pulling it out again. This is what Miqt mostly does internally.

examples/customsortfiltermodel/main.go
Ahuge: So I took a deeper look at this.

I can't figure out how to make this work as I would expect.

The only way I can get it working is with the following hack:
```go
dialog.OnResizeEvent(func(super func(event *qt.QResizeEvent), event *qt.QResizeEvent) {
		dialog.ResizeWithQSize(event.Size())
                // Width minus dialog.ContentMargins Left/Right
		treeView.Resize(event.Size().Width()-20, treeView.Size().Height())
	})
```

But having written Qt before, obviously that is a messy hack.


Is there a possibility that the default Resize events aren't being propagated down to child widgets?

examples/customsortfiltermodel/model.go
Ahuge: Ok so the TreeItem is created in `PopulateData` and the rootItem keeps track of its children.
And we keep a reference to the `rootItem`.

I don't know if that is enough to keep it alive.

I am not super clear if pulling in a a pointer from the QModelIndex and then casting it to a TreeItem pointer is the right way to do that, I am open to another way to get the modelItem from a QModelIndex if there is one.

`item = (*TreeItem)(parent.InternalPointer())`
If I call `runtime.KeepAlive(item)` everytime that I get the InternalPointer, will that cause a memory leak?
Isn't `QModelIndex.internalPointer()` a go pointer anyways?

PR #182, opened 2025-03-03
README: Update installation commands for Qt libraries

rcalixte:


rcalixte:
I've finally gotten around to documentation and testing. We can add instructions for Arch-based distributions too if you'd like. For Qt 5 and non-Debian-based distros, I would just need to double-check the package names. Feel free to modify this PR to your liking.

PR #187, opened 2025-03-31
Update charts6

rcalixte:
* Update the example and package to be consistent


rcalixte:
Almost perfect. 😔

mappu:
Not the first time an example has broken :sweat_smile: and probably it won't be the last time unless we add CI to build all examples.

Issue #60, opened 2024-11-06
Add bindings for QML / Qt Quick


mappu:

mappu:
Some other bound libraries (e.g. QWebEngine) had references to Qml types. Currently, the binding omitted them because it's definitely possible to use QWebEngine from a widgets-only application without a Qml dependency.

But, those such functions should be exposed somehow for Qml users, maybe in another Go subpackage.

Issue #185, opened 2025-03-28
Build Flags?


rcalixte:
Would it make sense to add build flags to the library's Go code?

As an example, I'm thinking along the lines of:

```diff
diff --git a/qt6/gen_qapplication.go b/qt6/gen_qapplication.go
index 6c5cdecf..eb665e1f 100644
--- a/qt6/gen_qapplication.go
+++ b/qt6/gen_qapplication.go
@@ -1,3 +1,5 @@
+//go:build qapplication
+
 package qt6
 
 /*

diff --git a/qt6/gen_qwidget.go b/qt6/gen_qwidget.go
index ce0a27ba..d83c5283 100644
--- a/qt6/gen_qwidget.go
+++ b/qt6/gen_qwidget.go
@@ -1,3 +1,5 @@
+//go:build qwidget
+
 package qt6
 
 /*

```

The classic build tags are deprecated for the minimum version of Go supported and slated for removal upstream but we can append them too:
```go
// +build
```

This has the benefit of allowing client applications to specify specific libraries for inclusion in compilation which should reduce the compilation time and binary size for resulting applications. The drawback that I can see is that this adds the dependency of something like a Makefile or basic script to execute this:

```make
build:
    go build -tags "qapplication qwidget"
```

Among the other drawbacks is that if `go build` is executed instead of the Makefile and without the build tags, the compilation steps would be the same as they are now. There is also the work to generate the precise list per application and maintain it. Over time, this shouldn't be as much of an issue I think. The other potential issue is if there are duplicate tags (the same class exists for two different libraries and they are both in the bindings), then there is no way to deduplicate them and both would be included with the specific tag. This feels like a minimum factor but it could possibly be addressed in the binding code if it were to become a problem down the road.

If this is a desired change, I can open a pull request with the changes, update the README (add a question to the FAQ? create a new section?), and add a basic Makefile for each existing example. Each would be a separate commit for easier review. 😅

mappu:
> The drawback that I can see is that this adds the dependency of something like a Makefile or basic script to execute this:

One idea is to do the inverse e.g. `go build -tags disable_qapplication` as an optimization for people who need it, then the default stays as it is.

A big problem is that there is a huge amount of overlap between Qt classes - many class methods refer to other classes, etc.

- Either these would have to be large categories (e.g. qwidget would bring in every type referenced in a qwidget method including QMouseEvent, QDragMoveEvent, QFont, QLayout just to name a few..) or, 
[...]

PR #189, opened 2025-04-02
 Add Qt 6 Qml

VelorumS:
I'm not sure about the use of QNetworkAccessManager* by Qml. Looks like just a pointer, not a full dependency on Networking. Maybe there is an easy way to fix it?

And more generally:

How defining QML types from Go would look like?

How to build QML modules for those types? (the CMake `qt_add_qml_module()`)

Other things in this PR:
- helloqml6 example
- gitignore vscode
- Manjaro install commands


mappu:
This looks fantastic. Thank you for contributing!

Fixes: #60 

I think blocking the networking types within Qml is the right move for now. I expect the way to make it work is to parse the qml package after the networking package, so it can detect types and set up the imports, but that does mean importing qml will always cause importing network. The current package structure isn't really set up for partial/optional dependencies: this has also been an issue for e.g. QWebEngine, which has some minor dependencies on QML types, creating a package dependency that you otherwise don't want.

The next move is probably splitting the headers into two with build tags (#185) or moving some into a qml_and_network/ directory, ... 

> How defining QML types from Go would look like?

@arnetheduck explained it to me once, it involves defining a custom qMetaObject and doing the equivalent of a runtime `moc` for it, but I think this will require a lot more work to expose the type out to the Qml side.

> How to build QML modules for those types? 

`qt_add_qml_module` in cmake packs the .qml source code into a .qrc resource bundle. You can use qrc/rcc resources with miqt, see miqt-rcc, but it's probably easier to `go:embed` the qml file and use `QQmlApplicationEngine::loadData` to load it.

mappu:
Works on my machine. :thumbsup: 

screenshot

The `qt6-declarative-dev` is sufficient at compile-time, but in order to run this example Debian packages all the runtime qml modules separately, so i'm just going to push an extra commit for the README.

arnetheduck:
seaqt indeed already exposes QML - there's a few more things to add to the block list to ensure everything compiles:

https://github.com/seaqt/seaqt-gen/blob/542f61c57a7a35f7e964e28f1d120b9a0ff6174c/cmd/genbindings/config-allowlist.go#L618 

A useful test is to compile all generated c++ files, to catch any such stragglers (if this was part of the Makefile, it could easily be added to CI as well).

Some of these are from the opengl parts that you ultimately also must wrap / support to get coverage for the QML api:

```
pkg-config --libs Qt6Quick
-lQt6Quick -lQt6QmlMeta -lQt6QmlWorkerScript -lQt6QmlModels -lQt6OpenGL -lQt6Gui -lQt6Qml -lQt6Network -lQt6Core
```

Binding the Qml library allows you to more or less open existing qml files and inject trivial values. That's not much, because most qml apps eventually need to create QObjects that QML consumes - for that, you indeed need to be able to expose meta-object description - not sure about go, but in languages with sufficient compile-time capabilities (ie c++11+, nim, rust etc) such things are done with templates/macros but can also be done with a codegen. There's also the option to do this dynamically (a bit like qt-dbus does it), but you lose out on some features. https://github.com/seaqt/seaqt-gen/commit/f279db8616501282a36a2874670532acc8cc4c2b exposes some of the plumbing needed for this (metacall and staticMetaObject above all).

You also need access to some private data when generating these - https://github.com/seaqt/seaqt-gen/commit/755c203ff85f3206da11d976023685977764514b is a trick to get access - in fact, this `void*` trick is broadly useful in wrappers for any private type because often you just need to pass an existing pointer around (happens both in qml and qmetaobject) without understanding it.

The other thing you'll find yourself needing soon is the ability to connect signals for the objects you've exposed - here, you need access to some private qt headers (the same that Qml uses, incidentally): https://github.com/seaqt/seaqt-gen/blob/542f61c57a7a35f7e964e28f1d120b9a0ff6174c/libseaqt/libseaqt.cpp#L17

I'd love to see the above things backported to miqt, but they all require someone to look at the corresponding go code as well.

VelorumS:
> A useful test is to compile all generated c++ files, to catch any such stragglers

Ok, I'll need to do at least that for this PR.

Exposing Go objects to Qml is a separate story. The oldest Go qml.v1 library looks the best from the user's point of view. From looking at that and seaqt-gen, it's probably possible to develop those bindings.

VelorumS:
I'm not seeing any more Qml issues with the generated c++ files. Other stuff like https://github.com/seaqt/seaqt-gen/blob/542f61c57a7a35f7e964e28f1d120b9a0ff6174c/cmd/genbindings/config-allowlist.go#L618 probably comes not from Qml, but from QtQuick which isn't useful at this stage.

arnetheduck:
> but from QtQuick 

Indeed - for future (backporting?) reference, I extracted the pieces needed for Quick as well as Qml on v5 here, rebased on top of the work done in this branch: https://github.com/seaqt/seaqt-gen/commit/b671983314e8e3de68d5d796b085d3d3dad645e1.

PR #195, opened 2025-04-10
Qt 6.9: Return type for QObjectData_dynamicMetaObject is now const

### LINKED ISSUE #194: Qt 6.9 not compiling - QObjectData_dynamicMetaObject  invalid conversion from const
When trying to compile the webEngine example I'm getting an error:

```go
package main

import (
	"os"

	"github.com/mappu/miqt/qt"
	"github.com/mappu/miqt/qt6/webengine"
)

func main() {
	qt.NewQApplication(os.Args)
	w := webengine.NewQWebEngineView2()
	w.Load(qt.NewQUrl3("https://www.github.com/mappu/miqt"))
	w.Show()

	qt.QApplication_Exec()
}
```

my go.mod

```
module github.com/emad-elsaid/example

go 1.24.2

require github.com/mappu/miqt v0.9.0 // indirect
```

The error I'm getting:

```
➜ go run  -ldflags '-s -w' .
# github.com/mappu/miqt/qt6
gen_qobject.cpp: In function ‘QMetaObject* QObjectData_dynamicMetaObject(const QObjectData*)’:
gen_qobject.cpp:40:39: error: invalid conversion from ‘const QMetaObject*’ to ‘QMetaObject*’ [-fpermissive]
   40 |         return self->dynamicMetaObject();
      |                ~~~~~~~~~~~~~~~~~~~~~~~^~
      |                                       |
      |                                       const QMetaObject*
```

tried with dynamic linking:
```
CGO_ENABLED=1 go run -ldflags="-linkmode=external" .
# github.com/mappu/miqt/qt6
gen_qobject.cpp: In function ‘QMetaObject* QObjectData_dynamicMetaObject(const QObjectData*)’:
gen_qobject.cpp:40:39: error: invalid conversion from ‘const QMetaObject*’ to ‘QMetaObject*’ [-fpermissive]
   40 |         return self->dynamicMetaObject();
      |                ~~~~~~~~~~~~~~~~~~~~~~~^~
      |                                       |
      |                                       const QMetaObject*
```

rcalixte:
* Closes #194

### DIFF
--- a/qt6/gen_qobject.cpp
+++ b/qt6/gen_qobject.cpp
@@ -36,7 +36,13 @@ void miqt_exec_callback_QObject_disconnectNotify(QObject*, intptr_t, QMetaMethod
 } /* extern C */
 #endif
 
+// This method's return type was changed from non-const to const in Qt 6.9
+#if QT_VERSION >= QT_VERSION_CHECK(6,9,0)
+const QMetaObject* QObjectData_dynamicMetaObject(const QObjectData* self) {
+#else
 QMetaObject* QObjectData_dynamicMetaObject(const QObjectData* self) {
+#endif
+
 	return self->dynamicMetaObject();
 }
 
--- a/cmd/genbindings/config-allowlist.go
+++ b/cmd/genbindings/config-allowlist.go
@@ -651,6 +651,10 @@ func ApplyQuirks(className string, mm *CppMethod) {
 		mm.BecomesNonConstInVersion = addr("6.7")
 	}
 
+	if className == "QObjectData" && mm.MethodName == "dynamicMetaObject" {
+		mm.ReturnType.BecomesConstInVersion = addr("6.9")
+	}
+
 	if className == "QFileDialog" && mm.MethodName == "saveFileContent" && mm.IsStatic {
 		// The prototype was changed from
 		// [Qt 5 - 6.6] void QFileDialog::saveFileContent(const QByteArray &fileContent, const QString &fileNameHint = QString())
--- a/cmd/genbindings/emitcabi.go
+++ b/cmd/genbindings/emitcabi.go
@@ -790,6 +790,14 @@ extern "C" {
 
 `)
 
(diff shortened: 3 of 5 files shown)


rcalixte:
The macro is copied but shouldn't represent a licensing issue. I added the comment so that it would be clear. Feel free to modify it as needed.

qt6/gen_qobject.cpp
```diff
@@ -36,7 +36,13 @@ void miqt_exec_callback_QObject_disconnectNotify(QObject*, intptr_t, QMetaMethod
 } /* extern C */
 #endif
 
+// This method's return type was changed from non-const to const in Qt 6.9
+#if QT_VERSION >= QT_VERSION_CHECK(6,9,0)
```
arnetheduck, line 40: can technically return the const pointer for all qt versions

rcalixte: Isn't this a type mismatch if the underlying type isn't `const`? And couldn't it lead to undefined behavior on Qt versions under 6.9?

Ultimately, this is a stop-gap of sorts. Once we start building on a version at or above Qt 6.9, the patches in `config-allowlist.go` can go away. (It probably makes sense to leave the remaining code... just in case. 😅)

arnetheduck: > Once we start building

An interesting approach would actually be to generate the bindings for each qt version separately and use the `#if` trick for any "new" features - ie one would start with 6.0 generating all of that, then gradually add each version as it happens, hidden behind `#if` - then the mappings could / would work with any qt version that the user happens to have installed.

> Isn't this a type mismatch

you can assign a non-const pointer to a const pointer of the same underlying type without issue, [in general](https://en.cppreference.com/w/cpp/language/implicit_conversion#Qualification_conversions) - the other direction is not allowed though.

mappu: As well as running genbindings on every version of Qt and diffing/merging the results, the other step would be to run it for clang-linux, clang-mingw, and clang-osxcross to identify all the cross-platform functions (#59).

Issue #188, opened 2025-04-02
Unstable name mangling


arnetheduck:
The current name mangling scheme for disambiguating overloads simple appends a number to the method name (New2, New3 etc).

While simple, it has a number of problems:

* when calling the function, it's hard to tell which is which - in particular, with types such as `QVariant` that have overrides for "similar" types, one might pick the wrong number by accident via features like integer size promotion (ie calling QVariant_newXX with a short might compile with the `int` override as well)
* between qt5/6 but also in minor feature releases, new overloads may be added - these cause renumbering if added in the middle - again, QVariant is a good example where the `QVariant_new6` takes a `QDataStream` or `long long` depending on the version of qt
* adding support for new types such as [`QVariantList`](https://github.com/seaqt/seaqt-gen/commit/b11a46733d5a4a870f809ce4b6a1ba8d37201176) again causes renumbering, as can be seen in the linked code where 22 and onwards get shifted by the new overload

Two potential ways out:

* parameter names - ie `QVariant* QVariant_new4(int i);` becomes `QVariant* QVariant_new_i(int i);` - this is pleasant to read but runs the risk of further conflicts and remains semantically unstable since variable names don't matter in C++ overload resolution rules and therefore might change as part of "cleanups"
* type names - `QVariant* QVariant_new4(int i);` becomes `QVariant_new_int` - the more correct approach, though slightly ugly for pointers, references and consts which have to be mangled as well
  * _probably_ most such qualifiers can simply be stripped, except for cases where two types would collapse to the same name such as overloads for both const and non-const or `*`/`&`

mappu:
> Two potential ways out: parameter names ... type names

There is a limited implementation of this already for function overloads (see transformoverload.go).

But when writing applications, I've found these overloads actually less ergonomic than the numbered versions. The longer names really make the code appear to deviate more from Qt C++ which makes porting more difficult. With numbers, it's less clear to read, but it's easy to write; usually you hit the autocomplete shortcut in your IDE, scroll to the one with the right-looking parameters, and move on with the project - 

Another idea to affix the stability (without addressing usability at all) would be to emit the ordering rules as some sort of file format (xml/json) and commit that for future runs of the generator. Then later blacklisting/new function additions/... could preserve the existing ordering as a matter of configuration.

arnetheduck:
> There is a limited implementation 

[...]

PR #196, opened 2025-04-11
Assorted minor breaking changes

mappu:
Updates: #188

- Move quirk handling to a standalone AST pass instead of inline in clang2il
- Treat the LinuxOnly quirk the same way as other quirks
    - This uncovered some more methods using LinuxOnly types
- Show `pkg-config` error messages in the genbindings output
- Fix invalid generation of any virtual methods returning `QString*`
- Treat optional parameters the same way as overloads for name resolution _(BREAKING)_
    - This fixes some "22", "23" etc method numbering to use a consistent, single set of numbers
- Remember whether a type was QMap/QHash and emit the correct C++ class import
- (by @arnetheduck) Allow QVariantList _(BREAKING)_
    - The QVariant constructor compatibility break is softened by sorting the list to the end. The remaining impact is limited to only QStateMachine/QSignalEvent
- (by @arnetheduck) Change C ABI integer mappings
    - Waiting to see if CI passes on mingw LLP64


arnetheduck:
re integer mappings, the correctest thing to do _iff_ the generated code should use the stdint types is actually to generate code like this:
```c
void func(qint64* p) {
   // TODO nil check
   int64_t tmp = static_cast<int64_t>(*p);
   qtfunc(*tmp);
   *p = static_cast<qint64>(tmp);
}
```

This applies in particular to >= 32-bit types which have a zoo of options and where code like above might be needed to not run afoul of strict aliasing rules. I did not have time to dig into the issue fully however.

mappu:
The part of the integer pointer aliasing i'd really like to solve, is the part in `emitcabi.go` `emitCABI2CppForwarding` -

```golang
// CABI has these as int64_t* (long int) which fails a static_cast to qint64& (long long int&)
// Hack a hard C-style cast
```

I think this commit is probably enough to remove the qint64 and qlonglong cases, but there are still a couple other dodgy ones in there.

cmd/genbindings/transformctors.go
```diff
@@ -4,30 +4,56 @@ import (
...
 	// QFoo(QWidget* parent);
-	checkIsDefaultCtor := func(candidate CppMethod) bool {
+	checkIsDefaultCtor := func(candidate *CppMethod) bool {
 		return len(candidate.Parameters) == 1 &&
 			candidate.Parameters[0].ParameterType == "QWidget" &&
 			candidate.Parameters[0].Pointer
 	}
 
+	// QFoo(QVariantList);
+	checkIsQVariantListCtor := func(candidate *CppMethod) bool {
```
arnetheduck, line 18: hairy :grimacing:

arnetheduck: ie the renumbering same happens for some `connect` overloads, private types, `void*` parameters etc that eventually are needed for example for metaobject support.

mappu: Yes, of course. This doesn't fix the issue in general, only, mitigates some of the impact of this specific change. If we come up with a general solution it can supersede this one.

Issue #193, opened 2025-04-07
Add bindings for qtermwidget


mappu:
qtermwidget powers the terminal widget in QTerminal. It is somewhat like libvte.
 
- https://github.com/lxqt/qtermwidget
- Debian package: libqtermwidget5-1
- GPL license

PR #198, opened 2025-04-12
Add libqtermwidget bindings

mappu:
Updates: #193 

These bindings are not working yet.

Debian 12 only has the Qt 5 version of this library. Some work needs to be done to the genbindings docker container to hybridize what packages it has available before the Qt 6 version can be added.

Based on the branch in #196, that should land first.


e1z0:
I have tested current status of the qtermwidget bindings on macOS, it runs well after some patching, here are the repo if anyone is interested: https://github.com/e1z0/miqt-qtermwidget it's a bit hacky approach but it works perfectly.

PR #203, opened 2025-04-18
webchannel/cflags.go: Add Qt6WebChannelQuick to the pkg-config

rcalixte:
This should aid the resolution for QQmlWebChannel

This should mean that this class can be moved back into the parent directory and the code can be removed, hopefully simplifying long-term maintenance.

### DIFF
--- a/docker/genbindings.Dockerfile
+++ b/docker/genbindings.Dockerfile
@@ -14,6 +14,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
         qt6-declarative-dev \
         qt6-multimedia-dev \
         qt6-svg-dev \
+        qt6-webchannel-dev \
         qt6-webengine-dev \
         libqscintilla2-qt5-dev \
         libqscintilla2-qt6-dev \
--- a/qt6/webchannel/cflags.go
+++ b/qt6/webchannel/cflags.go
@@ -1,6 +1,6 @@
 package webchannel
 
 /*
-#cgo pkg-config: Qt6WebChannel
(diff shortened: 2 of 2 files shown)


mappu:
I think this is related to the Qt version - it's [not in Qt 6.4 in Debian Bookworm](https://packages.debian.org/search?suite=bookworm&arch=any&mode=path&searchon=contents&keywords=Qt6WebChannelQuick.pc) but it [is present in Qt 6.8 in Debian Trixie](https://packages.debian.org/search?suite=trixie&arch=any&mode=path&searchon=contents&keywords=Qt6WebChannelQuick.pc).

I don't know exactly when the split happened, nor whether this is just a Debian thing or an upstream Qt thing. But I guess in order to support a broad range of Qt 6 versions, we're forced to split up the packages manually.

rcalixte:
> I don't know exactly when the split happened, nor whether this is just a Debian thing or an upstream Qt thing. But I guess in order to support a broad range of Qt 6 versions, we're forced to split up the packages manually.

Could this wait until there's a 6.8 branch and go exclusively in that branch? I don't see any Debian-specific patches for this (and it's an issue on Arch too) so I assume it was done upstream.

arwarw:
I just stumbled upon the same problem in NixOS, and applying this proposed PR fixes `go build qt6/webchannel/qmlwebchannel` there (version qtwebchannel-6.11.1).

As far as I can see, this is an upstream change:
https://github.com/qt/qtwebchannel/commit/48d2066d6c9b38f29368e5205a5b7f49ab3fa857
From the git history, this seems to be a change that was made in dev and is in 6.6.0-beta1 and later releases.

PR #205, opened 2025-04-20
Add Qt PDF

### LINKED ISSUE #200: Add binding for QPdfView
Just a wishlist item if it isnt a problem.

rcalixte:
* Closes #200


rcalixte:
I'm fairly certain that the use of the icons in the example PDF files falls under fair use but I can add a README or otherwise modify them if you're concerned about that. It wasn't easy to think of a good example PDF. 😅

PR #220, opened 2025-05-09
replace "localhost/" prefix when searching for images to support Podman

mrex:


mappu:
Thank you! This change looks great.

Issue #122, opened 2025-01-04
Missing QTimer.OnTimeout signal


mappu:
https://doc.qt.io/qt-6/qtimer.html#timeout

> Note: This is a private signal. It can be used in signal connections but cannot be emitted by the user.

djmattyg007:
It also looks like this constructor can't be used: https://doc.qt.io/qt-6.5/qtimer.html#singleShot-1

PR #224, opened 2025-05-21
lcdclock6: Add Qt 6 lcdclock example

rcalixte:
Following up on #223, I thought it would be good to have an example that covered a few new concepts (while also being a little fun). This features a QTimer example along with styling via Qt Style Sheets and QLCDNumber, QFont, and QTime.


mappu:
Looks great. Thank you!

Still need to figure out #122 at some point but I'm glad the QTimer subclass event can be used as an easy alternative. The other way is to use a Go timer and jump between Go/Qt threads, also less convenient.

rcalixte:
> Still need to figure out #122 at some point but I'm glad the QTimer subclass event can be used as an easy alternative. The other way is to use a Go timer and jump between Go/Qt threads, also less convenient.

It's on my list for post-6.8 things but I suspect we'll need to create a virtual method for private signals that will call `QObject::connect` directly and then take a function pointer or callback as a parameter. The user-facing API would then just have a single `On` method that maps to that.

rcalixte:
> It's on my list for post-6.8 things but I suspect we'll need to create a virtual method for private signals that will call `QObject::connect` directly and then take a function pointer or callback as a parameter. The user-facing API would then just have a single `On` method that maps to that. Thinking about it a little bit more, we would probably have a new field in the `CppClass` struct type named `PrivateSignals` that we would iterate over to do all of this.

I've confirmed it on the back-end that this works. I can open a pull request for that piece now if that works for you or I think I can get the front-end implemented by the end of this weekend and then update this example to use `OnTimeout` so that we know it all works correctly too. I'm fine with whatever you prefer.

I dumped the methods to verify the data structures were correct, including potential parameters:

```text
QTimer: [ {timeout{void false false 0 false false<nil><nil>}[] false false true false false false false false[] false<nil>} ];

QTimeLine: [ { valueChanged{void false false 0 false false<nil><nil>}[{x qreal false false 0 false false<nil><nil>}] false false true false false false false false[] false<nil> } { frameChanged{void false false 0 false false<nil><nil>}[{param1 int false false 0 false false<nil><nil>}] false false true false false false false false[] false<nil> } { stateChanged{void false false 0 false false<nil><nil>}[{newState QTimeLine::State false false 0 false false<nil><nil>}] false false true false false false false false[] false<nil> } {finished{void false false 0 false false<nil><nil>}[] false false true false false false false false[] false<nil>} ];
```

PR #225, opened 2025-05-24
Add private signals

rcalixte:


mappu:
Updates: #122 

This is fantastic work, thank you. I've learned something today about QPrivateSignal :thumbsup: 

It's really great to have the onTimeout signal - in Qt/C++ this is definitely the more normal way to use QTimer rather than subclassing TimerEvent. It will make porting applications easier.

I do see what you mean about the duplication. The generator code is starting to have a few different parts like this that are mostly similar, but, refactoring can be done separately later.

This almost completely solves #122 - the remaining part was the QTimer::singleShot static method, that must have some other cause for not being generated.

rcalixte:
> This almost completely solves #122 - the remaining part was the QTimer::singleShot static method, that must have some other cause for not being generated.

I looked into this. It's a functor parameter that is causing the current block. I have that on my list to tackle after we sort out the Clang issues blocking 6.8 and get 6.8 released. It ended up on my list as part of the QMetaType::registerConverter unblocking but there are a decent amount of mostly static methods that take some form of a function as a parameter. We'll get there...

arnetheduck:
For functors, a slot connection helper that turns a C function pointer + environment into a c++ functor (and allows releasing memory on disconnection) is needed, like so: https://github.com/seaqt/seaqt-gen/commit/55f6615add25b7b6fdeebb2e7e8150b06abe5d5a#diff-b87fe76f1cc1944e77611360b27220578882c5c90b5cf67455b7167542fb5b9a

PR #226, opened 2025-05-25
genbindings: Streaming parse clang output to reduce memory, get correct filepaths

arnetheduck:
To prepare genbindings for accepting arbitrary libraries from the command line, we must retain parse types not only from the header being processed but also its includes, to understand things like type hierarchies etc without depending on a particular module order.

This is the first baby step that addresses clang AST filtering.

The change has some collateral benefits: the improved accuracy of the AST parsing keeps more relevant information but at the same time reduces the memory footprint since filtering is done streaming instead of loading everything into memory first (ditto when writing the cache).

A consequence is that a lot more clang processes can run in parallel without OOMing.

* parse "file" correctly to discover type provenance - using the standard go json parser for this does not work since the  depend on serialization order which go discards (see
https://github.com/dtolnay/clang-ast?tab=readme-ov-file#source-locations)
* stream clang output to JSON decoder and stream-write the cache to reduce memory footprint
* with the newfound memory efficiency, bump up the number of parallel jobs and use threads for the parsing as well
* fix missing classes such as QSysInfo using the corrected `file` source location field
* don't try to re-run clang if it fails (OOM is unlikely anyway)


arnetheduck:
draft because of the numbered renames - need some help with these in case  we want to merge this

cmd/genbindings/clangast.go
```diff
@@ -0,0 +1,341 @@
...
+	return fmt.Sprintf("0x%x", uint64(h))
+}
+
+type AstNode struct {
+	Id    HexUint64  `json:"id,omitempty"`
+	Kind  string     `json:"kind"`
+	Inner []*AstNode `json:"inner,omitempty"`
+	// TODO is there a simple way to "flatten" the fields so they appear as in
+	// the original AST?
+	Fields map[string]interface{} `json:"fields,omitempty"`
```
mappu, line 54: I think the only way is to type out the struct definition in full.

It has the downside that sizeof(AstNode) consumes space for empty values for your other types, so ideally you want polymorphic JSON decode into a lot of different AstNode specialized types, which is somewhat annoying to achieve with encoding/json alone.

arnetheduck: yeah, I had that thought .. miqt only uses a small subset of fields so the overhead is not necessarily that big compared to the instantiation of a map for every node and it would be more convenient to "use" at the expense of making clangast miqt-specific. maybe that's the better tradeoff (vs keeping this module generic / non-miqt-specific)

qt6/qml/gen_qqmllist.h
```diff
@@ -29,13 +29,11 @@ typedef struct QVariant QVariant;
 #endif
 
 QQmlListReference* QQmlListReference_new();
-QQmlListReference* QQmlListReference_new2(QVariant* variant);
```
arnetheduck, line 32: the interesting thing about these removed overloads is that `_3` and `_4` are the same .. :shrug:

qt/gen_qstringlist.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
```
mappu, line 1: The rest of the code sees QStringList as being a `KnownTypedefs` alias for `QList<QString>` to get a nice native type projection instead.

I guess it's showing up because of the `file` changes? Might be better to block in Widgets_AllowHeader(),

arnetheduck: [f85157b](https://github.com/mappu/miqt/pull/226/commits/f85157b52d813b02301b703d85119345547f8d36)

rcalixte:
Do we want to assess the conflicts this will have with #118? The thought of rebasing somewhat terrifies me. I don't know that I'll be applying this change downstream either.

rcalixte:
> Semantically, I believe there is no conflict

Handling `clang2il.go` and `main.go` seem to be the worst of it. It would help a little if you could revert the changes in `config-libraries.go` as they are just formatting fixes that conflict with the entire restructuring in the other PR. I also structured the commits in the other PR so that they are mostly single files (the exceptions being config-libraries.go & main.go since they are logically consistent and the rebuild of the library which is mostly trivial to rebase). If I'd gotten the other PR done in time, we could've avoided this. Apologies. 😔

arnetheduck:
> revert the changes in config-libraries.go

[9d2f200](https://github.com/mappu/miqt/pull/226/commits/9d2f200f19e227048cae1a46c4000a2f2c9eebf4)

rcalixte:
> > revert the changes in config-libraries.go
> 
> [9d2f200](https://github.com/mappu/miqt/pull/226/commits/9d2f200f19e227048cae1a46c4000a2f2c9eebf4)

Thank you... and I hate to be a nag but this PR will likely not be squashed and merged which means rebasing still means processing each commit from the PR... meaning this adds to the rebase effort. Rather than nagging like this, would it be easier if we just got the other PR sorted first? Is there any rush on this being merged?

arnetheduck:
I don't mind either way, from the test merge I did locally of the two, there were only a few small adjustments regardless of the order. Conflicts happen, it's a fact of life in async development ;)

I have a few other PR:s lined up that could also maybe benefit miqt that depend on this one, so it's more a matter that it's holding those up as well and since yours is still in draft, I didn't want to wait indefinitely before posting this one.

arnetheduck:
> probably this is just the gunzipping

Nah, there's just more data to parse, in general, because each cached file now contains more data - basically, the full translation unit (including `#include`) instead of just the symbols from the one header file. unzip _usually_ is balanced out by the smaller read on disk but if you unzip all of them, you'll see that they've grown quite big.

After this PR it's possible to further refactor the code to process each header file out-of-order, ie all the globals in the generator can be turned to locals. 

In another language, one would probably use a "cache"/work area per thread to save on the analysis, not sure how that's done in golang since you don't control the threads - probably a "cache pool" of some sort where each "job" checks out a cache and hands it back when it's done.

With that out of the way, the header processing itself can be paralelised, but more importantly, you can run it on any one single header file and get a deterministic result.

Issue #228, opened 2025-05-26
Unable to get menu item icon to show using a designer based UI


greeneg:
Trying to add an icon with UI code that was generated via Designer, but cannot for the life of me figure out why the QIcon::ThemeIcon entries will not show up in the application. The main.go looks like so:

```golang
package main

import (
	"os"

	qt "github.com/mappu/miqt/qt6"
)

func main() {
	qt.NewQApplication(os.Args)

	// Create a new main window
	ui := NewMainWindowUi()
	ui.MainWindow.Show()

	qt.QCoreApplication_SetAttribute(qt.AA_DontShowIconsInMenus)
	qt.QApplication_Exec()
}
```

And a "simplified" output of the generated code looks like so:

```golang
// Generated by miqt-uic. To update this file, edit the .ui file in
// Qt Designer, and then run 'go generate'.
//
//go:generate miqt-uic -InFile maininterface.ui -OutFile maininterface.go -Qt6

package main

import (
	"os"

	qt "github.com/mappu/miqt/qt6"
)

type MainWindowUi struct {
	MainWindow             *qt.QMainWindow
	centralwidget          *qt.QWidget
	horizontalLayoutWidget *qt.QWidget
	horizontalLayout       *qt.QHBoxLayout
	folderTreeWidget       *qt.QTreeWidget
	menubar                *qt.QMenuBar
	fileMenu               *qt.QMenu
	helpMenu               *qt.QMenu
	statusbar              *qt.QStatusBar
	quitMenuEntry          *qt.QAction
	aboutMenuEntry         *qt.QAction
}

// NewMainWindowUi creates all Qt widget classes for MainWindow.
func NewMainWindowUi() *MainWindowUi {
	ui := &MainWindowUi{}
	ui.MainWindow = qt.NewQMainWindow(nil)
	MainWindow__objectName := qt.NewQAnyStringView3("MainWindow")
	ui.MainWindow.SetObjectName(*MainWindow__objectName)
	MainWindow__objectName.Delete() // setter copied value
	ui.MainWindow.Resize(880, 641)
	ui.MainWindow.SetWindowTitle("Folder Lister")

	ui.quitMenuEntry = qt.NewQAction()
	quitMenuEntry__objectName := qt.NewQAnyStringView3("quitMenuEntry")
	ui.quitMenuEntry.SetObjectName(*quitMenuEntry__objectName)
	quitMenuEntry__objectName.Delete() // setter copied value
[...]

Issue #215, opened 2025-05-05
miqt-uic: support more missing properties


mappu:
- [ ] QTextEdit.font
- [ ] QToolBar.toolBarBreak

pitkes22:
### QGraphicsView

- [ ] minimumSize
- [ ] maximumSize
- [ ] baseSize
- [ ] cursor

PR #229, opened 2025-05-27
miqt-uic: Add test fixtures, add reentrancy, fix QIcon::ThemeName

mappu:
Fixes: #228 

Updates: #215 

Minor improvements to the miqt-uic tool.

- Detect and patch the known issue with `QIcon::ThemeName` generation (#228)
- Add support for arbitrary connections
- Add support for tooltips
- Mark the top-level windowText for translation
- ~Fix QAction shortcuts being marked for translation~
- Support spacers
- Internal:
    - Refactor global variables into struct (resolves repeated use of `generate()` in test code)
    - Add some test cases for Qt 5, and some 6.4 / 6.9 cases at risk of differing


greeneg (commented):
Overall, looks good. I need to test this as a branch validation against my code to see what happens. I'll try to test this in the coming day, since after that, I'll be away from my dev rig until the 2nd of June

mappu:
I've added some files from Designer 6.9 showing the new QIcon::ThemeName behaviour.

I like the new theme icon picker, it's much nicer than the old text field. It seems like if the .ui file contains an new enum-name it appears in the top "Icon theme" dropdown, and if the .ui file contains an old-style name it uses the second "XDG theme" dropdown instead.

image

cmd/miqt-uic/ui2go.go
```diff
@@ -595,28 +688,32 @@ func generateWidget(w UiWidget, parentName string, parentClass string, useQt6 bo
...
+}
+
+func normalizeSignalName(s string) string {
+	if s == "" {
+		return ""
 	}
 
-	return ret.String(), nil
+	s = strings.TrimSuffix(s, `()`)
+	return strings.ToUpper(string(s[0])) + s[1:]
```
rcalixte, line 700: The signals aren't always empty, e.g.:
```
<signal>valueChanged(int)</signal>
```

Here's what I put into a similar method:
```suggestion
	if idx := strings.IndexByte(s, '('); idx != -1 {
		return s[:idx]
	}
	return s
```

mappu:
@greeneg I modified the helloworld6 example to load your supplied `maininterface.go` file, and, the icons do show up for me on Debian 12:

image

cmd/miqt-uic/ui2go.go
```diff
@@ -170,8 +207,13 @@ func renderProperties(properties []UiProperty, ret *strings.Builder, targetName,
 			customContentsMargins = true
 
 		} else if prop.StringVal != nil {
-			//  "windowTitle", "title", "text"
-			ret.WriteString(`ui.` + targetName + setterFunc + `(` + generateString(prop.StringVal, parentClass) + ")\n")
+			//  "windowTitle", "title", "text", "shortcut"
+
+			if prop.Name == "shortcut" { // Shouldn't be translatable
+				prop.StringVal.Notr = true
+			}
```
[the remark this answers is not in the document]
mappu: :eyes: You're right, C++ uic marks these for translation. miqt-uic should usually have the same behaviour.

cmd/miqt-uic/ui2go.go
mappu: Thanks for the snippet! I'll do it that way.

This will work as long as the signal and slot have the same number of arguments. It is possible for the signal to have more arguments than the slot, so it might need changes later.

greeneg:
Hmmm, k. Time to do some digging via strace to see if I can find out what is going on

PR #239, opened 2025-06-04
cabi: assorted changes

arnetheduck:
* use `std::nothrow` new operator, avoiding a cross-language exception propagation (the language projection can check for null to catch OOM)
* remove some spurious whitespace in generated code
* group header functions with newline
* reduce variable naming differences in emitBindingHeader/emitBindingCpp


mappu:
Looks good to me. std::nothrow is a nice addition and widely available in all compilers (even old GCC 5 in the Qt5 win32 container).

Go throws an uncatchable fatal on OOM, which is sort of the same as you get from a null pointer dereference, it's not a language in which it's possible to sensibly recover from an OOM condition, so I don't think the Go projection would really benefit from extra null pointer checks on the results.

PR #247, opened 2025-07-24
podman support, missing sudo condition

dezza:


mappu:
Thank you for the PR! There was some previous support for podman in #220, but i guess it required `DOCKER=podman` or a docker symlink present on the system.

cmd/miqt-docker/main.go
```diff
@@ -151,7 +151,7 @@ func getDockerRunArgsForGlob(dockerfiles []fs.DirEntry, containerNameGlob string
 		fullCommand = append(fullCommand, "-t")
 	}
 
-	if runtime.GOOS != "windows" {
+	if needsSudo && runtime.GOOS != "windows" {
```
mappu, line 154: A very common way to run docker on Linux is to add your regular user to the docker group. But then, containers default to running as UID=0 (or the `USER` from the Dockerfile) instead of as your own user. I wanted miqt-docker to, by default, act with the same permissions as if you had run `go build` commands natively.

Is it safe to leave this here for podman? Or does it cause problems?

cmd/miqt-docker/docker.go
```diff
@@ -22,24 +22,53 @@ type dockerImage struct {
...
+	_, err := exec.LookPath(cmd)
+	return err == nil
+}
+
+func findContainerBackend() string {
+	if cmdExists("docker") {
+		return "docker"
+	}
+	if cmdExists("podman") {
+	  return "podman"
```
mappu, line 35: Please run `gofmt` :smile:

cmd/miqt-docker/main.go
dezza: podman is rootless by default (and still runs as a fake root `UID=0` inside container in a high range of UIDs).

The equivalent would be: https://docs.docker.com/engine/security/rootless/#prerequisites.

If it poses problems, tell me how and I will correct - it would be fixed with a simple condition - I can also check for podman if you like.

So if you use

mappu: I tried this out and it works correctly in the needsSudo case. So - LGTM!

mappu: Moved to #268 .

PR #255, opened 2025-08-20
Add Qt SQL

rcalixte:


rcalixte:
Qt SQL is LGPL 3 licensed. Qt SQL also ships by default alongside Qt Core, GUI, Widgets, etc. in most (all?) cases so the documentation and the Dockerfiles should be fine as-is.

The example is a port of the example from Qt's example documentation [here](https://doc.qt.io/qt-6/qtsql-sqlwidgetmapper-example.html). The upstream example itself is 3-clause BSD licensed but this shouldn't matter for our purposes. The example attempts to create a SQLite database in-memory and in-process to use for the example and wires up the data model to the GUI, allowing for edits that save to the database for both the QLineEdit and QTextEdit fields. This also introduces the first usage of QDataWidgetMapper in the examples. While both Qt 5 and Qt 6 libraries generated fine, the Qt 6 QSqlDatabase driver on Debian 12 doesn't ship with any drivers by default so that example only failed (correctly though!) which is why I removed it. Aside from the imports and the window title, it's line-for-line the same with the Qt 5 example so it's extremely trivial to restore when that time comes.

One key thing: you'll notice on L84 of the example that I used QStyledItemDelegate instead of QSqlRelationalDelegate which inherits from it. This doesn't cause any issues for the example because the only usage is actually as their parent QAbstractItemDelegate in one instance as a parameter on the next line. QSqlRelationalDelegate is completely skipped in the generation of the bindings. This is because all of the functions defined in that class are FunctionDecls which we currently have marked TODO. I'm going to look into this. I see it sort of like the last example we had with private signals and the usage of QTimerEvent until we got OnTimeout properly added to the bindings. Once we get the FunctionDecl type properly supported in the library, we'll update the example to know that it's working properly. I didn't see anything in these dumps that looked useful for parsing either but I'm holding out hope I guess. In the worst case, maybe a manual typedef? My only concern is that this touches the same files as the other Clang PRs and I'm wary to dip into more conflicts. Let me know how you'd like to proceed here.

mappu:
:ship: The PR looks fantastic as usual. Thank you very much for sorting this out!

There are definitely a lot more clang issues to solve - it will be great once the current big two PRs are merged in. The fast AST parsing PR would currently strip unused fields like FunctionDecl, but that's a very easy conflict to solve - the bigger risk is if a newer Clang drops the field entirely or does something else strange,

PR #261, opened 2025-09-01
Add detailed info about building on macOS arm64

ksckaan1:
Related discussion -> https://github.com/mappu/miqt/discussions/257


mappu:
Thanks for the PR! I'm glad to get confirmation that this is working.

PR #267, opened 2025-09-30
Add WhereAmI to the app list

rubiojr:
Thank you for the amazing bindings!


mappu:
Thanks for using Miqt! Your app looks really nice.

Issue #268, opened 2025-10-03
miqt-docker: needsSudo detection fails to add --user if user does not need sudo


mappu:
> this actually breaks running with regular docker .. `needsSudo` is set only if some previous docker command fails but if the user running docker has docker permissions in general, the docker commands won't fail and as a result, docker will be run as root which means the generated files will be owned by root due to the missing --user flag

_Originally posted by @arnetheduck in https://github.com/mappu/miqt/pull/247#discussion_r2344245832_

PR #272, opened 2025-10-18
Docker: Add qt6.10, add MIQTDOCKER_UID, fix --user detection

mappu:
Fixes: #268 

- Add new container for Qt6.10 testing, based on debian13 + aqtinstall
- miqt-docker: Add new MIQTDOCKER_UID option to force the value of `--user` if autodetection fails
- miqt-docker: Fix wrong uid mapping decision on Linux if the current user can run docker without sudo
- miqt-docker: Fix wrong uid mapping decision on Docker for Mac

### DIFF
--- a/cmd/miqt-docker/main.go
+++ b/cmd/miqt-docker/main.go
@@ -154,36 +154,32 @@ func getDockerRunArgsForGlob(dockerfiles []fs.DirEntry, containerNameGlob string
 
 	// On Linux, Docker runs as root, and our generated files would be owned by 0:0
 	// Use `--user $uid:$gid` to avoid this
-	// This is not needed for podman, and it's not needed for Windows, and it's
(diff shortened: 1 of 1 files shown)


cmd/miqt-docker/main.go
```diff
@@ -151,13 +152,38 @@ func getDockerRunArgsForGlob(dockerfiles []fs.DirEntry, containerNameGlob string
 		fullCommand = append(fullCommand, "-t")
 	}
 
-	if needsSudo && runtime.GOOS != "windows" {
+	// On Linux, Docker runs as root, and our generated files would be owned by 0:0
+	// Use `--user $uid:$gid` to avoid this
+	// This is not needed for podman, and it's not needed for Windows, and it's
```
arnetheduck: it's not needed but is it harmful? why not always supply it by default and have an option to disable?

PR #280, opened 2025-10-28
miqt-docker: show docker's output instead of discarding it

ilius:
We really want to know what docker is doing, specially when it takes a long time to download stuff and when it gets stuck.


mappu:
Thanks! This makes sense to me. Most of the docker containers are pretty quick, but in particular the win64-cross-go1.23-qt6.8-static is really slow as it has to compile Qt from source.

PR #285, opened 2025-11-11
[EXPERIMENT] Run genbindings separately for qt5/qt6

mappu:
Experiment only, just to have something to talk about.

Related: https://github.com/mappu/miqt/discussions/284


arnetheduck:
added a variation on this in https://github.com/seaqt/seaqt-gen/pull/15 - turns out this is a convenient way to run the generator on multiple qt versions as well.

The next step is really to run the generator per qt module, so that it can be run separately for a project like scintilla (perhaps providing a `.json` file with things like blocklists etc that are specific to scintilla).

In miqt, the json approach is slightly more tricky because some qt modules are not 1:1 mapped (ie widgets and core are not separate modules in the binding), but it's probably not insurmountable - seaqt simply derives all relevant stuff from the "module name".

Issue #100, opened 2024-12-05
Translation


MatusOllah:
Hello,
I'm trying to implement translations in my project.
I see that miqt has methods for translating like `QTranslator_Tr` but I'm not sure how to generate the `.ts` files required for translation and how to properly integrate them into my project.

How do I translate properly in miqt?
Any examples or detailed instructions would be greatly appreciated!

mappu:
Thanks for your patience. There's not yet an easy answer.

In Qt C++ the process is:
1. Mark strings with `tr("something")`
    - It's a static class method, so that your C++ class name can be part of the "context" of the translation
2. Use `lupdate` to scan source code and extract .ts files
3. Translate .ts files with Qt Linguist
4. Add `TRANSLATIONS += file.ts` in the qmake .pro file.

In MIQT several steps will have to work differently:
1. Mark strings with `qt.QApplication_Tr("something")` or any other class version
    - Some short wrapper function would be really helpful 
2. Extract strings using some other extractor, probably [some xgettext implementation to .po format](https://github.com/gosexy/gettext)
3. Translate .po files with PoEdit
4. Convert .po files to .ts with `lconvert`
5. Use `lrelease` to pack .ts source to .qm binary file
6. Embed .qm binary file using either qrc/rcc process or singular go:embed process
7. Load the qm file at runtime: `translator := qt.NewQTranslator()`, `translator.load("file.qm")`, `qt.QCoreApplication_installTranslator(translator)`

This is not a good experience so it may be easier to stay within some pure-Go translation ecosystem. Only the downside is the qm translation system must be used to translate Qt system text (e.g. "OK"/"Cancel" buttons within default dialogs)

MatusOllah:
Alright, tried embedding the translations files using `miqt-rcc` and loading them using `QCoreApplication_InstallTranslator`, but it didn't work as expected. I think I'll stick with a pure Go i18n package instead.
Closing this issue as resolved for now.

Here’s my code for reference in case you find it helpful:
```go
package main

import (
	"log/slog"
	"os"
	"runtime"

	_ "github.com/MatusOllah/gcharted/assets"
	_ "github.com/MatusOllah/gcharted/i18n"
	"github.com/MatusOllah/gcharted/internal/gui"
	"github.com/MatusOllah/slogcolor"
	qt "github.com/mappu/miqt/qt6"
)

func main() {
	// Logger
[...]

PR #287, opened 2025-11-20
Add miqt-lupdate

rcalixte:
I'm working on implementing this for both downstreams and even though I saw #100, I figured that since it was trivial enough to port that it would be worth a pull request.

For the Qt 5 example, I used `tr` method variants and for the Qt 6 example, I used `QCoreApplication::translate` variants. I thought this would be effective for demonstrating both possible cases without too much redundancy. I opted not to include a README for each but I can add something if you think it's needed.

One thing is that I'm not usually a fan of committing binaries so if you want to rebuild the .qm files just as a matter of practice, that works for me. Qt Linguist or `lrelease` should be able to compile the translation source files as they are. I also opted not to use a .qrc to embed the translations for similar reasons. It's an extra step that is optional and I'm not sure adds much when there is already an example for that purpose. The English is untouched so that the banner is visible. By default, Qt Linguist removes any extra XML comments in the header.

For the program code, I tried to implement as much of the non-deprecated schema as possible so that imported translation files wouldn't be clobbered when being written back out.

There are still some things I want to polish up, mostly around handling some special cases but as is, the tool should work for most cases and especially for initializing translations.


mappu:
Wow - fantastic work! :partying_face: LGTM but I just have to fix up #289 first before hitting merge.

I think the small compiled `.qm` output files are fine to leave in the examples, they're not hiding anything.

mappu:
I have a couple questions that can totally be ignored and left for future work, not a blocker at all.

I had a quick look at the comment handling - have you tried adding comments into the Go source and seeing how they appear in the xml output and in Linguist? Usually this is something like:

 ```cpp
//: Hello translators, "Asdf" is the name of a thing
// secret secret, my second line comment is an internal reference, not extracted for translation
foo->SetText(tr("Asdf"));

//: "Fdsa" is the name of a different thing
bar->SetText(tr("Fdsa"))
```

It looks like lupdate [automatically detects `//:` as a magic comment prefix](https://doc.qt.io/archives/qt-5.15/i18n-source-translation.html#translator-comments), but other popular extractors like [xgettext](https://man7.org/linux/man-pages/man1/xgettext.1.html) will let you customize the matched comment prefix (`--add-comments=TRANSLATORS:` is somewhat widely used as being more explicit than a magic colon character).

Just from what I read, by comparison, the 2nd argument to `tr()` is a disambiguator if you have a homograph in the source language, i.e. both `tr("right", "correct")` and `tr("right", "direction")` where they would be explicitly translated differently and show up as two different entries in Linguist (e.g. :de: richtig, rechts).

The extractor tools also often let you define additional functions that should be extracted (e.g. xgettext `--flag=my_tr:3` will start to match the 3rd argument to function `my_tr` ; in lupdate this is `-tr-function-alias`).

rcalixte:
I've patched the disambiguation and handled comment cases for comments starting with `//~`, `//:`, and `/*:`. I updated the documentation but didn't call out `//~` since it isn't technically handled the same as upstream. (Should it just be removed then?)

mappu:
I think this is wonderful as-is. :tada: Thanks for doing the `//:` handling!

The only downside of the standard function names is they are just a bit long to type in a real program full of strings. In my experience a short alias or wrapper is often used to get closer to the ideal `_()` or `tr()`. From that perspective i'd go with "adding to the list" instead of "replacing the list".

But as-is today,  this all works _great_ and it catches all the miqt-uic autogenerated texts. More features can always be added in the future. :grin:

PR #295, opened 2025-12-16
Add Qt Positioning

rcalixte:
I aimed for a privacy-preserving approach to the examples here. The implementation for the examples is more or less the same but the backend patches to support each version of the library varied so the value of the similar examples is a bit hidden away but it's there. 😄 

One key question that I had was around the design of the callbacks. I know it would be a large/breaking change but I found the lack of the pointers to the object of the callback to require reworking the examples. Two QDoubleSpinBox objects and their callbacks could easily scale to many more and of other types in a more robust application and I think having to manage duplicated code at that scale could be a little unwieldy. Downstream, the pointers are passed and I used object names for each QDoubleSpinBox and checked the name in the callback to update accordingly. Here, I had to update both values in the callback even though just one changed and move both objects to the global scope. Some of this comes down to differences in implementation or coding style but using this example, if there was a third QDoubleSpinBox for the altitude, it would be the difference between duplicating the code block for the definition and tweaking the conditional in the callback (plus the associated string printing changes) or doing all of that plus having to duplicate the callback if the function was defined inline.

I don't know if this is a limitation due to the vtable implementation (thinking about inherited callbacks here) or if it even makes sense to implement. It was an unexpected bump in an otherwise seamless process so I thought it was at least worth bringing up. I'll cc @arnetheduck for any thoughts too. If it makes sense to move to a Discussion outside of the pull request (since the PR should be ready to go), let me know and I can start that too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ examples/libraries/extras-scintillaedit/extras-scintillaedit
 examples/libraries/qt-multimedia/qt-multimedia
 examples/libraries/qt-network/qt-network
 examples/libraries/qt-pdf/qt-pdf
+examples/libraries/qt-positioning/qt-positioning
 examples/libraries/qt-printsupport/qt-printsupport
 examples/libraries/qt-script/qt-script
 examples/libraries/qt-sql/qt-sql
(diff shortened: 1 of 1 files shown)


mappu:
Thanks again @rcalixte , this looks great as always!

---

I think Go doesn't generally need a context pointer in the callbacks since you can do it Go side - 

```golang
var coord *positioning.QGeoCoordinate
lon.OnValueChanged(func(value float64) {
    // access coord as a local variable directly, through Go heap-escaping magic
})
```

or with currying - 

```golang
func barbeque[T1, T2 any](car T1, cdr func(T1, T2)) func(T2) {
	return func(cons T2) {
		cdr(car, cons)
	}
}

func main() {
    // [...]
    var coord *positioning.QGeoCoordinate
    lon.OnValueChanged(barbeque(coord, onValueChanged))
    // [...]
}

func onValueChanged(coord *positioning.QGeoCoordinate, value float64) {
    // same implementation as per example, but coord is now available as a parameter variable
}
```

I can definitely see the mapping of multiple signals to one slot becoming more unwieldy, as you add more dimensions, but that's just a code organization issue. I think the example as-is is perfectly clear and shows that the library works :+1:

PR #301, opened 2025-12-30
Add Qt6 QStateMachine

krya-kryak:
Closes #599

Works as expected, includes an [upstream example](https://doc.qt.io/qt-6/qtstatemachine-trafficlight-example.html) ported by @rcalixte


mappu:
This looks fantastic, thank you @krya-kryak ! 

The CI pipeline verifies that the bindings can be identically generated from source, so everything looks good there.

- We should test that it works - can you add a mini test app into the examples/libraries/... directory? It doesn't have to do anything useful, just instantiate a class and see if it compiles OK.
- There's a minor merge conflict because another library was also added to `config-libraries.go`, it should be easy to resolve

mappu:
One more thing - right now genbindings.Dockerfile is unable to compile the new Go package subdirectory because it doesn't know what CFLAGS to use - you need to manually add a `cflags.go` file to each of the new directories, that has the pkg-config statement in it for the new libraries.

Check out any other subdirectory for an example, they all have one,

rcalixte:
It looks like I spoke too soon. The generation for SCXML is failing due to the generation of [Loader::load](https://doc.qt.io/qt-6.5/qscxmlcompiler-loader.html#load). The pointer to the QStringList and in a virtual function no less in particular is troublesome. Our options in the short-term are to try to implement it, block just the method, or remove the entire SCXML module (since I don't know if it can work without the Loader class). If you only need State Machine, the latter two are the quickest options to at least get the PR into good shape. I know SCXML is related and the installation package is the same but given that it is not part of the Qt 5 binding, maybe it is okay to remove for now?

rcalixte:
@krya-kryak One more and I think we're actually in the clear this time: https://github.com/krya-kryak/miqt/pull/2. :)

mappu:
Thanks both of you for your hard work! Everything here looks fine to me.

PR #305, opened 2026-01-19
fix `noexcept` parsing

arnetheduck:
`noexcept` optionally comes with a condition, which is sometimes
filled in - a prominent example being the QDateTime constructor.

[backported](https://github.com/seaqt/seaqt-gen/commit/0eb5d3688592e89676bcaeb37b25f1748c6da3d0) from seaqt


mappu:
Wow, interesting find. How did you run into this?

Thanks for updating the unit tests too,

Looking over the diff, the end-user changes that resulted here are
- QCommandLineParser, QWebEnginePage/QWebEngineView: more methods
- QStyleOptionXxx: add many new operatorAssign() methods
- QItemSelection, QVersionNumber: new ctor (no churn)
- Qt 5 QDateTime: new ctor (**churn**)

Since the library now has real users, I wonder if it's worth hacking another sort order override in transformctors.go to force the new 0-argument ctor to the end of the list...? Or probably not worth bothering since it's only Qt 5...

arnetheduck:
> How did you run into this?

By coming across code that wanted to create a datetime ;) afair are more cases than the ones listed but miqt is not able to parse them for other reasons, this is just a stepping stone.

> if it's worth hacking 

I've since moved on to [mangle names](https://github.com/seaqt/seaqt-gen/pull/7) in a different way, avoiding this problem space entirely. Ostensibly one could generate both styles, though that's probably not of interest for seaqt at least.

Notably, every time you fix a "ErrTooComplex" to parse a new C++ construct, this problem will likely repeat.

> now the ctor order has to differ

I ran across a case like this for QVariant which was pretty hard to track down, because it ended up actually compiling and calling the wrong constructor (qvariant constructors change quite a bit between 5 & 6) - part of the reason for the name mangling switch.

mappu:
I'd be more than happy to adopt that style of C mangling, it definitely would be more stable - it's just the Go representation that's the real problem, since Go doesn't have overloading.

Go can only fake it by boxing everything in `...any` and figuring it out at runtime, but that seems awful

PR #319, opened 2026-03-03
Add pixels examples

rcalixte:
These examples make liberal use of components from the Graphics View framework, including events, leading to a fun application to play with.

cc @5k3105 Mostly ported this example from the other Go bindings with a few tweaks. Fortunately, much of what appear to have presented issues with the other library were seamless here.


5k3105:
You fixed the memory spikes with Delete() - well done!

rcalixte:
> You fixed the memory spikes with Delete() - well done!

Those are explicit but not strictly necessary since the library makes use of finalizers for directly constructed types.

mappu:
image

PR #320, opened 2026-03-16
Add .gitattributes

rcalixte:
This is a cosmetic change to have the header files properly identified as C rather than the current C++ to accurately reflect the project's structure.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.h linguist-language=C


mappu:
LGTM. Theoretically the same header files are parsed as both C by CGO and C++ when building the cpp translation units, but if you consider C to be a subset of C++ then this is definitely a safer move.

It would also be possible to mark all the `gen_` files as `linguist-generated`, i think we discussed this previously but it had some pros and cons, and i don't exactly recall the outcome / can't find the discussion about it.

rcalixte:
> LGTM. Theoretically the same header files are parsed as both C by CGO and C++ when building the cpp translation units, but if you consider C to be a subset of C++ then this is definitely a safer move.

This shouldn't affect anything with the library itself. It's merely a cosmetic change for the files in the repository.


> It would also be possible to mark all the `gen_` files as `linguist-generated`, i think we discussed this previously but it had some pros and cons, and i don't exactly recall the outcome / can't find the discussion about it.

I vaguely remember this back when the forward declarations were initially added. That was the change that I think flipped the language graph?

PR #328, opened 2026-05-08
Add screenshot examples

rcalixte:
This is ported from the [upstream example](https://doc.qt.io/qt-6/qtwidgets-desktop-screenshot-example.html). The only caveat is that it uses QScreen which, on Linux, only supports X11. For proper Wayland support, I think we would have to re-implement it some day using widgets from Qt Multimedia. Maybe a fun exercise in the future. 😅


mappu:
:ship: Thanks again for the awesome work!

---

Yeah - no image under Wayland - 

image

With `QT_QPA_PLATFORM=xcb` to start under XWayland it works, but i guess XWayland only offers it a blank screen:

image

Looks like upstream bug [QTBUG-34976](https://qt-project.atlassian.net/browse/QTBUG-34976). Also [[1]](https://qt-project.atlassian.net/browse/QTCREATORBUG-29815) is probably the same underlying problem.

I've used [robotgo](https://github.com/go-vgo/robotgo) before, but last i looked, their screenshot also only supported X11 (and Windows, etc).

rcalixte:
I briefly looked into gating the Linux and Wayland conditional with an implementation that should support Wayland. I'm not closed off to implementing it some day because I think some folks could find value in a working example but it would bring in a library dependency so maybe it would make more sense as a separate example in the multimedia libraries directory?

PR #332, opened 2026-05-14
mainthread: add missing C++ standard flags

### LINKED ISSUE #331: Bug: qt/mainthread missing `#cgo CXXFLAGS: -std=c++11`
### Problem

`qt/mainthread/mainthread.go` has `#cgo pkg-config: Qt5Core` but does not set `#cgo CXXFLAGS: -std=c++11`. This causes compilation failures when the system C++ compiler defaults to a standard older than C++11 (e.g. MXE's MinGW GCC in the `win64-cross-go1.23-qt5.15-static` miqt-docker environment).

The `qt` package itself correctly sets this in `qt/cflags.go`:
```go
/*
#cgo CXXFLAGS: -std=c++11
#cgo CFLAGS: -std=gnu11
#cgo pkg-config: Qt5Widgets
*/
```

But CGo directives don't propagate across package boundaries, so `qt/mainthread` compiles `mainthread.cpp` without `-std=c++11`, and Qt's `qtypeinfo.h` uses C++11 features that fail to parse:

```
qtypeinfo.h:319:20: error: expected '}' before 'bool'
 Q_DECLARE_TYPEINFO(bool, Q_PRIMITIVE_TYPE);
```

### Reproduction

```bash
miqt-docker win64-cross-go1.23-qt5.15-static go build github.com/mappu/miqt/qt/mainthread
```

### Fix

Add `#cgo CXXFLAGS: -std=c++11` to `qt/mainthread/mainthread.go`:

```go
/*
#cgo CXXFLAGS: -std=c++11
#cgo pkg-config: Qt5Core

#include "mainthread.h"
*/
```

The same issue may affect `qt6/mainthread` as well.

### Environment

- miqt v0.13.0
- miqt-docker `win64-cross-go1.23-qt5.15-static`
- MXE GCC 11.2 (x86_64-w64-mingw32.static-g++)
- Qt 5.15.2

MarSoft:
The mainthread packages compile mainthread.cpp which includes Qt headers, but unlike qt/cflags.go they do not set the C++ standard. CGo directives don't propagate across package boundaries, so when the system C++ compiler defaults to an older standard (e.g. MXE's MinGW GCC), Qt headers fail to parse.

Add -std=c++11 for Qt5 and -std=c++17 for Qt6, matching the minimum requirements of each Qt version.

Reproducible with:
  miqt-docker win64-cross-go1.23-qt5.15-static \
    go build github.com/mappu/miqt/qt/mainthread

Fixes #331

### DIFF
--- a/qt/mainthread/mainthread.go
+++ b/qt/mainthread/mainthread.go
@@ -6,6 +6,7 @@ import (
 )
 
 /*
+#cgo CXXFLAGS: -std=c++11
 #cgo pkg-config: Qt5Core
 
 #include "mainthread.h"
--- a/qt6/mainthread/mainthread.go
+++ b/qt6/mainthread/mainthread.go
@@ -6,6 +6,7 @@ import (
 )
 
 /*
+#cgo CXXFLAGS: -std=c++17
 #cgo pkg-config: Qt6Core
 
 #include "mainthread.h"


mappu:
:+1: LGTM. Thanks for the quick fix!

We have the same -std ... in qt/cflags.go but interestingly, not in qt6/cflags.go. It could be added there too, but none of the current containers or known environments use Qt 6 with a compiler that does not already default to C++17 standards.

PR #334, opened 2026-05-24
Add Qt 5 Qwt

### LINKED ISSUE #294: Add binding for QOpenGLWidget
I have an application that used QOpenGLWidget and used QPainter to draw into it.

rcalixte:
This library has a plugin based on Qt UI Plugin that adds support for Qt Designer/Creator so it's good that we already have bindings for the UI Plugin components needed.

There were lots of examples to choose from, I chose sysinfo since it seemed about the right size but also pretty fun to see come to life.

I implemented a workaround for inheritance that originally started downstream because of QCustomPlot but then ended up expanding.

Due to the inheritance of some Qt OpenGL-based widgets, the bare minimum of Qt OpenGL was added, but it should not be directly used or considered full. On top of that, tons of blocks were implemented instead of extending support. I did make a small example that showed that they could be used directly albeit limited but I didn't want to encourage usage with an example. I think this at least closes #294.

This is only Qt 5 because the Qt 6 version won't be available in Debian until Debian 14. (I can't wait since this library quickly approached my favorite when compared to QCustomPlot and KPlotting, though those are also fine libraries.)

The license for Qwt is a modified version of LGPL 2.1 where the key differences are highlighted at the top of the license (mostly around attributions) but I don't think the license modifications rise to the level of being marked as restricted. I'll follow your lead here.


mappu:
Awesome! :partying_face: Works on my PC:

image

---

It's extremely exciting to see the start of potential OpenGL support. There is a lot more that can be done at a later point (e.g. a GLint is an int typedef, that would unlock a lot of methods immediately).

In genbindings, already the Scintilla libraries without convenient Debian packages are installed from source, that is an option for Qwt-qt6 if you don't want to wait until mid-2027 for Forky. Then at least the bindings would be available if people do have appropriate packages on their other distro.

I can see Debian packages too for a lot of other indirect things - libqwtplot3d, libqwtmathml...

Reading over the Qwt license - the modifications are granting additional permissions, so overall this is more permissive even than the LGPL, I think it definitely doesn't count as "restricted". But regarding the directories, I think the qt/ and qt6/ directories should be kept for official Qt libraries, and qt-extras may be better for permissive 3rd party libraries.

rcalixte:
> There is a lot more that can be done at a later point (e.g. a GLint is an int typedef, that would unlock a lot of methods immediately).

I burnt myself out a bit getting this done downstream for a project I'm working on. There are pointer use cases for GLint and the others too so enabling the type would mean having to handle the types like the other integer pointer parameter types. If there's enough interest, maybe we can slowly peel this back?



> In genbindings, already the Scintilla libraries without convenient Debian packages are installed from source, that is an option for Qwt-qt6 if you don't want to wait until mid-2027 for Forky.

I can take a look at replicating this setup.

> I can see Debian packages too for a lot of other indirect things - libqwtplot3d, libqwtmathml...

Oh... these are probably going to be added. 😅

> Reading over the Qwt license - the modifications are granting additional permissions, so overall this is more permissive even than the LGPL

I took the attribution portion to be stricter but other than that, agreed.

> qt-extras may be better for permissive 3rd party libraries.

I'll get this revision done too.

rcalixte:
> I can see Debian packages too for a lot of other indirect things - libqwtplot3d, libqwtmathml...

The 3D plotting library has headers but will need more care to patch properly. There is at least an example on the front page. The MathML doesn't appear to contain any headers.

All of the other revisions for this should be all set.

mappu:
Fantastic as always!

PR #340, opened 2026-07-01
Construct pure virtual methods in inherited classes

### LINKED ISSUE #246: Meaning of 'can only override virtual methods for directly constructed types'
Hi, currently giving a try to miqt to port legacy gtk2 code to qt6 on ubuntu 22.04. Everything goes well, it compiles and get executables. But as soon as I try to do something less trivial than the miqt provided examples, I run into problems. Simple example: I built `examples/modelview_color6`, it runs ok. Then I changed the code from

```
		return 1000
	})

```
to

```
	model.OnColumnCount(func(parent *qt.QModelIndex) int {
		return 2
	})
		return 1000
	})

```
Just added a method to have more columns and all what I get is:

`panic: miqt: can only override virtual methods for directly constructed types`

Can someone explain me where the problem lies?

rcalixte:
This unfortunately ends up as a sort of breaking change. Now that the base implementations are defined, they are returning 0 in some cases. The result is that not setting an override means a 0 value being returned for something like the column count, leading to an empty-looking model where Qt's private default may return 1 (in the case of QAbstractListModel but not QAbstractTableModel). The fix is mechanical but it's still a breaking change. I've updated the examples that I could find that were impacted.

While it's not strictly marshalling in the purest sense, the tests are for validation that can otherwise only be seen visually. Would it make sense to split the tests into a new testing nomenclature?

* Closes #246 

cc @arnetheduck


rcalixte:
I think I'm going to close this in favor of #341. I believe these methods are not meant for overrides and that we just need to address the panics in order to resolve #246.

PR #348, opened 2026-08-05
Detect and project default constructors

rcalixte:
I borrowed from your PR to move the default constructor loop to after the inner loop.


mappu:
Thanks as always @rcalixte !

Out of the three versions (this, #346, and https://github.com/5k3105/miqt/commit/f5c646a3187879cba3b65b562450b55bcdb269cf ) I think your version is the simplest and cleanest (and you were the first to get a passing CI run too, haha).

Issue #79, opened 2024-11-15
Project simple Qt types (QRect/F, QSize/F, QRgba64) as plain Go structs


mappu:
In Qt C++ these POCO-like types are normally stack-allocated and passed by value reference. But MIQT currently has to move them to heap pointers to cross the C ABI, requiring extra CGO roundtrips to allocate them from the Qt side.

By using plain Go structs instead, they can be better stored on the stack.

PR #349, opened 2026-08-11
Add tetrix examples

rcalixte:
* https://doc.qt.io/archives/qt-5.15/qtwidgets-widgets-tetrix-example.html

This started out without much reason but then it became a little more! The examples are a bit larger than I think is ideal but they do flow mostly logically well and this does introduce QRandomGenerator.

This example was the first time I got bit doing `defer object.Delete()` for manual cleanups. It happened with the QColor and QRect objects so I removed those `defer` calls. A reminder course on the fact that closures can heap allocate also led to some minor refactors. The Qt 5 version still has some sporadic garbage collection crashes that I haven't been able to resolve or even consistently trigger. Other than that, everything mapped over cleanly.

A fair warning: it starts out enjoyable... and then... you understand why there is no win state implemented. 😅


mappu:
I see what you mean about not needing a win condition... :sweat_smile: It's pretty addictive! When are you getting a Steam release?

No crashes in a few games on Qt 5 or Qt 6 - but the changes in #338 might have helped with the original problem?

For QColor and QRect, they're such simple types, i really do think that something like #79 and copying all the fields by value would be better than trying to make sure we don't mix up any heap memory. But, it's a good stress test at least.

---

Qt 6 screenshot

image

Qt 5 screenshot

image

rcalixte:
> No crashes in a few games on Qt 5 or Qt 6 - but the changes in #338 might have helped with the original problem?

This was written against master but the most recent refactor I did appears to have stabilized things. I was going to edit the message but I hadn't umm... tested... as many samples. I've gotten a lot of practice of late. 😄 

> For QColor and QRect, they're such simple types, i really do think that something like #79 and copying all the fields by value would be better than trying to make sure we don't mix up any heap memory. But, it's a good stress test at least.

I've thought about this but my issue is that we potentially lose all of the class member functions unless we implement it differently from all of the other native types. I'm actually in the middle of another example that uses QRgb and that's been enlightening as well. I assume you would want to include QPoint(F) and QSize(F) too? There's probably more. We would need to remove the constructors and destructors most likely. Maybe it would be a good exercise worth trying a hand-binding first to get the idea and see what it might take...

PR #353, opened 2026-09-02
Wrap around some enum values

rcalixte:
This reverts #115 and provides an alternate solution for #3. There are some unfortunate cases of pointers to enums as parameters and a different bit size means a different pointer alignment. This calculates two's complement for signed integers rather than storing the overflows as new types.


mappu:
Nice, that's a much cleaner solution than the _64bit files. I might only suggest we generate a `// overflow` comment as well in those cases but that's not particularly important.

:ship:

rcalixte:
> I might only suggest we generate a `// overflow` comment as well in those cases but that's not particularly important.

This is a small change that I should be able to get done. 👍🏼

PR #354, opened 2026-09-03
Add comments for overflowing enums

rcalixte:
Very much a TIL about Go's formatter here. 😅


mappu:
Huh, that is fascinating with the extra newlines.

Thanks for adding this! I just think it makes the negative numbers a bit more clear in the output.

